PR #248, opened 2025-04-25
persisting the message in conversations independent of AI service, fallback message added

rish231294:
…llback message added


rish231294:
Included changes with https://github.com/pipeshub-ai/pipeshub-ai/pull/265.

PR #343, opened 2025-05-11
Removed failed only check for re index button in connector stats

vishwjeet-pawar03:


vishwjeet-pawar03:
New updated pr made with uniform connector stats and sync events handling

PR #509, opened 2025-08-08
added graph interface and arango impl

rish231294:


rish231294:
all the comments in the pr are incorp.

PR #577, opened 2025-08-21
Bugfix/pdf render

rishimalhan:


rishimalhan:
My bad. this should have been on my fork.

PR #579, opened 2025-08-21
[ISSUE-578] PDF render bugfix

rishimalhan:
Context:
Running a VM instance on google compute engine and started pipeshub using dev docker compose.
Using Ngrok to forward port 3000 on a doman url
Google workspace syncs correctly and AI query goes through with a response and citations

Issue:
Upon viewing the citation I see Error loading pdf.
I sent backend logs to cursor and it came up with the following explanation

Cursor Output:
Perfect! I found the issue. The streaming route does exist in the knowledge base routes, but it's mounted at the wrong path. Looking at the code:
// In kb.routes.ts - this route exists: router.get( '/stream/record/:recordId', authMiddleware.authenticate, metricsMiddleware(container), ValidationMiddleware.validate(getRecordByIdSchema), getRecordBuffer(appConfig.connectorBackend), );

But in your app.ts, the knowledge base routes are mounted at /api/v1/knowledgeBase, so the full path becomes:
/api/v1/knowledgeBase/stream/record/:recordId
But the frontend is calling:
/api/v1/stream/record/:recordId
The Fix
You need to add a separate route for streaming at the root level. Update your app.ts:

### DIFF
--- a/backend/nodejs/apps/src/app.ts
+++ b/backend/nodejs/apps/src/app.ts
@@ -50,6 +50,7 @@ import { registerStorageSwagger } from './modules/storage/docs/swagger';
 import { CrawlingManagerContainer } from './modules/crawling_manager/container/cm_container';
 import createCrawlingManagerRouter from './modules/crawling_manager/routes/cm_routes';
 import { MigrationService } from './modules/configuration_manager/services/migration.service';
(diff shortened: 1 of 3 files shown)


vishwjeet-pawar03:
Hi @rishimalhan the implementation of this route already exists in the kb.routes.ts file, and if a connector url is set then it calls directly the python backend. Happy to help with any further queries.

PR #2439, opened 2026-05-28
fix(mcp): add sessionIdGenerator to StreamableHTTPServerTransport

berkabot:
## Problem

The MCP controller creates a `StreamableHTTPServerTransport` with `sessionIdGenerator: undefined`, which means the server never includes a `Mcp-Session-Id` header in its responses.

`mcp-remote` (the standard stdio↔HTTP bridge used by Claude Desktop and other MCP clients) requires this header to correlate responses. Without it, the connection hangs indefinitely on `initialize` and Claude Desktop times out.

## Fix

Pass `sessionIdGenerator: () => randomUUID()` so every session gets a unique ID returned in the `Mcp-Session-Id` response header. The server remains effectively stateless (no session storage needed) while satisfying the MCP Streamable HTTP spec and `mcp-remote` expectations.

## Tested

With this fix, the `Mcp-Session-Id` header appears in the response and Claude Desktop connects successfully via `mcp-remote`.

### DIFF
--- a/backend/nodejs/apps/src/modules/mcp/controller/mcp.controller.ts
+++ b/backend/nodejs/apps/src/modules/mcp/controller/mcp.controller.ts
@@ -6,6 +6,7 @@
  */
 
 import { Response, NextFunction } from 'express';
+import { randomUUID } from 'crypto';
 import { StreamableHTTPServerTransport } from '@modelcontextprotocol/sdk/server/streamableHttp.js';
 import { AuthenticatedUserRequest } from '../../../libs/middlewares/types';
 import { Logger, getLogLevel } from '../../../libs/services/logger.service';
@@ -40,7 +41,7 @@ export const handleMCPRequest =
       const { PipeshubCore } = await coreModule;
 
       const transport = new StreamableHTTPServerTransport({
-        sessionIdGenerator: undefined,
+        sessionIdGenerator: () => randomUUID(),
       });
 
       const { server: mcpServer } = createMCPServer({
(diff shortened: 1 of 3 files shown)


realaashil:
Hi @berkabot , I hope you are doing well, can you share me your claude desktop setup of  pipeshub mcp to futher debug this issue.

abhishek-pipeshub:
@berkabot Can you fix the unit tests failures as well

realaashil:
Hi @berkabot , I hope you are doing well. We diagnoised and found out that claude desktop doesnt support http transport protocol for mcp. SessionID is MAY (optional) according to mcp spec sheet and we want to align our mcp server with the spec. We support stdio transport out of the box and it supports oauth token (client_creds only). So you dont need to generate token yourself, I am pasting an example config below try it once and let me know your results or any issues faced. I will document this properly on our docs website later this week / next week.
```
"mcpServers": {
    "pipeshub": {
      "command": "npx",
      "args": [
        "-y",
        "@pipeshub-ai/mcp",
        "start",
        "--server-url",
        "http://localhost:3000/api/v1",
        "--client-id",
        "CLIENT_ID",
        "--client-secret",
        "CLIENT_SECRET",
        "--token-url",
        "http://localhost:3000/api/v1/oauth2/token"
      ]
    }
  },
```
  Thank you for your patience and have a great day!

realaashil:
Hi @berkabot , you can refer to latest docs here: https://docs.pipeshub.com/mcp/claude-desktop

PR #2531, opened 2026-06-14
fix(indexing): stop dropping the first table data row in LLM enhancement

rachit367:
## Summary

`_enhance_tables_with_llm` silently drops the **first data row** of every table when generating per-row natural-language descriptions, which shifts every row's description by one and leaves the **last row** with no description. This corrupts table retrieval/citation accuracy for any table with a header row and ≥2 data rows.

## Root cause

In `backend/python/app/events/processor.py`, header rows are filtered out before the grid is built:

```python
for i, (row_dict, row_block) in enumerate(zip(row_dicts, row_blocks)):
    is_header = (... table_row_metadata.is_header)
    if not is_header:
        non_header_row_dicts.append(row_dict)   # <-- grid is data rows only
        non_header_row_indices.append(i)

if non_header_row_dicts:
    table_data = {"grid": [[row] for row in non_header_row_dicts]}
    row_descriptions, _ = await get_rows_text(
        self.config_service, table_data, table_summary, column_headers   # <-- bug
    )
```

But `get_rows_text` assumes **`grid[0]` is the header row** and skips it when `column_headers` is non-empty (`indexing_helpers.py`):

```python
if column_headers:
    table_rows = table[1:]     # drops grid[0]
else:
    table_rows = table
```

Since the grid here has *no* header row, `table[1:]` removes the first **data** row. Result:
- `row_descriptions` has `N-1` entries for `N` data rows.
- The update loop maps `row_descriptions[k]` → `non_header_row_indices[k]`, so row 0 gets row 1's description, row 1 gets row 2's, …, and the last row is skipped by the `description_idx < len(row_descriptions)` guard.

The contract is confirmed by the other callers — `pymupdf_opencv_processor.py`, `docling_doc_to_blocks.py`, and the Azure processor all pass the **full** grid (header at `grid[0]`) to `get_rows_text`, so `table[1:]` is correct for them. This call site is the only one that pre-strips headers.

## Fix

Pass `[]` for `column_headers` at this call site, since the grid already contains data rows only. One-line change; `get_rows_text` and the other callers are untouched.

## Testing

Added a regression test in `TestEnhanceTablesWithLlm` (1 header + 3 data rows) asserting each data row keeps the description generated from **its own** position and the last row is labeled.

```
pytest tests/unit/events/test_processor.py::TestEnhanceTablesWithLlm
6 passed
```

Verified the test **fails on the previous code** (it received `column_headers=['Name','Role']` and dropped a row) and **passes with the fix** — so it genuinely guards the regression. Pure unit test, no DB/Kafka/Qdrant/LLM (the LLM helpers are mocked).

### DIFF
--- a/backend/python/app/events/processor.py
+++ b/backend/python/app/events/processor.py
@@ -786,9 +786,17 @@ class Processor:
 
                     if non_header_row_dicts:
                         try:
-                            table_data = {"grid": [[row] for row in non_header_row_dicts]}
+                            data_rows = [
+                                row_blocks[idx].data.get("cells", [])
+                                for idx in non_header_row_indices
+                            ]
+                            if column_headers:
+                                grid = [column_headers] + data_rows
+                            else:
+                                grid = data_rows
+                            table_data = {"grid": grid}
                             row_descriptions, _ = await get_rows_text(
-                                self.config_service, table_data, table_summary, []
+                                self.config_service, table_data, table_summary, column_headers
                             )
 
                             # Update row blocks with LLM descriptions (only non-header rows)
--- a/backend/python/app/events/processor.py
+++ b/backend/python/app/events/processor.py
@@ -788,7 +788,7 @@ class Processor:
                         try:
                             table_data = {"grid": [[row] for row in non_header_row_dicts]}
                             row_descriptions, _ = await get_rows_text(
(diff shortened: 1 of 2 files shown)


abhishek-pipeshub:
Fixed in different pr

PR #2677, opened 2026-07-06
fix(deployment): stabilize MongoDB failures

shekharkadyan:
## Summary

This is a focused MongoDB deployment stability bugfix.

On smaller Docker hosts, MongoDB's WiredTiger engine can reserve a large share of available memory by default. In the full PipesHub compose stack, that can push MongoDB into an OOM crash loop and then the services depending on MongoDB never become healthy.

There is also a separate host-specific failure mode where MongoDB 8.x can exit with `139` on newer kernels because of a glibc rseq/tcmalloc interaction.

## What Changed

- Cap MongoDB's WiredTiger cache with `MONGO_CACHE_GB` defaulting to `1` GB.
- Add a default MongoDB memory limit of `2G` in the main compose file so MongoDB cannot consume the whole Docker host.
- Keep the MongoDB image tag overridable through `MONGO_IMAGE_TAG` for recovery cases.
- Preserve the official MongoDB image default `glibc.pthread.rseq=0` for normal hosts, while allowing affected hosts to set `MONGO_GLIBC_TUNABLES=glibc.pthread.rseq=1` without editing compose files.
- Apply the same MongoDB runtime settings to the default, build-from-source, production, and integration compose entrypoints.
- Document the recovery and resource knobs in `env.template`.
- Wire the MongoDB compose regression test into PR validations when Docker Compose deployment files change.

## Why These Defaults

The defaults are based on the quickstart/self-hosted target: a 16 GB-class shared Docker host running MongoDB alongside the PipesHub app, Qdrant, graph DB, Redis, and optional broker services.

MongoDB WiredTiger's upstream default can reserve several GB on that kind of host because it sizes cache from available memory. That behavior is good for a dedicated MongoDB box, but risky for this compose stack because the other services need the same RAM. The `1` GB cache cap prevents MongoDB from taking a disproportionate share of memory, and the `2G` container limit leaves extra room for MongoDB process overhead, connections, journaling, and spikes above the cache.

These are stability-first defaults, not maximum-performance MongoDB tuning. On larger or dedicated hosts, raise `MONGO_CACHE_GB` and `MONGO_MEMORY_LIMIT` together. Avoid decreasing below `1` GB cache unless this is a very small dev workload.

## Performance Notes

The new defaults intentionally favor small-host stability. On larger dedicated hosts, the `1` GB WiredTiger cache cap can reduce MongoDB cache hit rate for very large or Mongo-heavy workloads. The setting is overridable with `MONGO_CACHE_GB`, and the main compose memory limit is overridable with `MONGO_MEMORY_LIMIT`.

For the expected quickstart/self-hosted profile, this is a reasonable tradeoff because PipesHub also runs the app, vector DB, graph DB, Redis, and optional broker services on the same machine. Capable hosts can raise the limits without editing compose.

## Automated Tests

- [x] CI: `.github/workflows/main.yml` now has a `validate-docker-compose` job in `PR Validations`.
  - It runs automatically on PRs when `deployment/docker-compose/**` or `.github/workflows/main.yml` changes.
  - It executes `bash deployment/docker-compose/tests/mongodb_stability_test.sh`.
  - It skips unrelated PRs so normal backend/frontend changes do not pay for compose validation.
- [x] Unit-style config test: `bash deployment/docker-compose/tests/mongodb_stability_test.sh`
  - Renders all affected compose files as JSON.
  - Asserts MongoDB defaults to `mongo:8.0.17`.
  - Asserts the WiredTiger cache cap is present and defaults to `1` GB.
  - Asserts `GLIBC_TUNABLES` defaults to `glibc.pthread.rseq=0` and remains overridable.
  - Asserts `MONGO_IMAGE_TAG`, `MONGO_CACHE_GB`, `MONGO_GLIBC_TUNABLES`, and `MONGO_MEMORY_LIMIT` overrides render correctly.
- [x] Integration/config validation:
  - `docker compose -f deployment/docker-compose/docker-compose.yml --env-file deployment/docker-compose/env.template config --quiet`
  - `docker compose -f deployment/docker-compose/docker-compose.build.neo4j.yml --env-file deployment/docker-compose/env.template config --quiet`
  - `docker compose -f deployment/docker-compose/docker-compose.integration.arango.yml --env-file deployment/docker-compose/env.template config --quiet`
  - `docker compose -f deployment/docker-compose/docker-compose.integration.neo4j.yml --env-file deployment/docker-compose/env.template config --quiet`
  - `docker compose -f deployment/docker-compose/docker-compose.prod.yml --env-file deployment/docker-compose/env.template config --quiet`

## Manual Tests Recommended

- [ ] On a smaller host or Docker Desktop VM, start the default compose stack and confirm MongoDB remains healthy instead of entering an OOM restart loop.
- [ ] On a host that reproduces MongoDB `exit 139`, set `MONGO_GLIBC_TUNABLES=glibc.pthread.rseq=1` in `.env`, restart MongoDB, and confirm it stays healthy.
- [ ] On a larger host, optionally raise `MONGO_CACHE_GB` and `MONGO_MEMORY_LIMIT`, then run `docker compose config` before restarting to verify the intended values render.
- [ ] If using the fallback image pin, set `MONGO_IMAGE_TAG` intentionally and verify the rendered compose config uses that tag before starting. Do not downgrade an existing MongoDB 8.x data volume to 7.x without wiping the volume.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -151,3 +151,34 @@ jobs:
           else
             echo "No Python files changed in backend/python, skipping pyright."
           fi
+
+  validate-docker-compose:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Check changed deployment files
+        id: changes
+        run: |
+          git fetch origin main
+          MERGE_BASE=$(git merge-base origin/main HEAD)
+          CHANGED_FILES=$(git diff --diff-filter=d --name-only "$MERGE_BASE" -- \
+            'deployment/docker-compose/**' \
+            '.github/workflows/main.yml' || true)
+
+          if [ -n "$CHANGED_FILES" ]; then
+            echo "changed=true" >> "$GITHUB_OUTPUT"
+            echo "Docker Compose validation will run for:"
+            echo "$CHANGED_FILES"
+          else
+            echo "changed=false" >> "$GITHUB_OUTPUT"
+            echo "No Docker Compose deployment files changed, skipping validation."
+          fi
+
+      - name: Validate MongoDB compose stability settings
+        if: steps.changes.outputs.changed == 'true'
+        run: bash deployment/docker-compose/tests/mongodb_stability_test.sh
--- a/deployment/docker-compose/docker-compose.build.neo4j.yml
+++ b/deployment/docker-compose/docker-compose.build.neo4j.yml
@@ -186,11 +186,14 @@ services:
       start_period: 300s
 
   mongodb:
-    image: mongo:8.0.17
+    image: mongo:${MONGO_IMAGE_TAG:-8.0.17}
     restart: always
+    # Keep MongoDB from reserving half the host RAM and allow the rseq workaround.
+    command: ["mongod", "--bind_ip_all", "--wiredTigerCacheSizeGB", "${MONGO_CACHE_GB:-1}"]
     environment:
       - MONGO_INITDB_ROOT_USERNAME=${MONGO_USERNAME:-admin}
       - MONGO_INITDB_ROOT_PASSWORD=${MONGO_PASSWORD:-password}
+      - GLIBC_TUNABLES=${MONGO_GLIBC_TUNABLES:-}
     volumes:
       - mongodb_data:/data/db
     healthcheck:
--- a/deployment/docker-compose/docker-compose.build.neo4j.yml
+++ b/deployment/docker-compose/docker-compose.build.neo4j.yml
@@ -193,7 +193,7 @@ services:
     environment:
       - MONGO_INITDB_ROOT_USERNAME=${MONGO_USERNAME:-admin}
       - MONGO_INITDB_ROOT_PASSWORD=${MONGO_PASSWORD:-password}
-      - GLIBC_TUNABLES=${MONGO_GLIBC_TUNABLES:-}
+      - GLIBC_TUNABLES=${MONGO_GLIBC_TUNABLES:-glibc.pthread.rseq=0}
     volumes:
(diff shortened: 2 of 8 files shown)


shekharkadyan:
Addressed Gemini review comments in `ec045913`:

- `GLIBC_TUNABLES` now defaults to `glibc.pthread.rseq=0` across all MongoDB compose entrypoints, preserving the official image default and TCMalloc per-CPU cache performance on normal hosts.
- `MONGO_GLIBC_TUNABLES=glibc.pthread.rseq=1` remains available as the opt-in workaround for hosts where MongoDB exits 139.
- `env.template` now documents `MONGO_CACHE_GB` and `MONGO_MEMORY_LIMIT` alongside the Mongo recovery knobs.
- `mongodb_stability_test.sh` now asserts the rendered default `GLIBC_TUNABLES` value.

Validated with `bash deployment/docker-compose/tests/mongodb_stability_test.sh` and `docker compose config --quiet` for all affected compose files.

shekharkadyan:
Added rationale for the MongoDB resource defaults in `60d4d4c8`.

The code comments and PR body now explicitly state that `MONGO_CACHE_GB=1` and `MONGO_MEMORY_LIMIT=2G` are stability-first defaults for a 16 GB-class shared quickstart host running MongoDB alongside the app, Qdrant, graph DB, Redis, and optional broker services. They also note that larger/dedicated hosts should raise both values together, while decreasing below 1 GB cache is only appropriate for very small dev workloads.

shekharkadyan:
Added automatic CI coverage in `f47872ac`.

The MongoDB compose stability test is now part of `.github/workflows/main.yml` under a new `validate-docker-compose` PR validation job. It runs automatically when a PR changes `deployment/docker-compose/**` or the workflow itself, and skips unrelated PRs.

abhishek-pipeshub (approved):

PR #2685, opened 2026-07-08
fix: prevent spreadsheet indexing from storing LLM-rewritten row facts

TatshSiow:
Problematic Approach, Ignore this


TatshSiow:
This approach is problematic, mark as invalid, will do another research for the issue

PR #2722, opened 2026-07-16
increased batch size & concurrency for embedding creation

kaushal1999:


abhishek-pipeshub:
Need to test this properly

PR #2836, opened 2026-07-29
feat(connectors): add Zendesk connector for tickets, Help Center, users, groups and organizations

anupPradhan0:
# 🚀 Add Zendesk Connector

## Overview

This PR introduces a **fully functional Zendesk connector** for PipesHub.

Previously, Zendesk existed only as a placeholder connector. It appeared in the UI but could not perform synchronization. This PR replaces that stub with a production-ready implementation that supports full and incremental synchronization, permissions, search indexing, and Help Center ingestion.

The implementation follows the existing Zammad connector architecture while taking advantage of Zendesk's native incremental export APIs.

---

# ✨ Features

### Supported Entities

| Entity | PipesHub Entity |
|---------|-----------------|
| Tickets | `TicketRecord` |
| Ticket Comments | Ticket blocks |
| Ticket Attachments | `FileRecord` |
| Help Center Articles | `WebpageRecord` |
| Help Center Sections | `RecordGroup (KB)` |
| Zendesk Groups | `RecordGroup (PROJECT)` |
| Organizations | `RecordGroup (USER_GROUP)` + `AppUserGroup` |
| Users (Agents & End Users) | `AppUser` |

---

# 🔧 What's Included

## Connector

- Full Sync support
- Incremental Sync using Zendesk Incremental Export API
- Cursor persistence using `SyncPoint`
- Ticket, Article, User, Group and Organization synchronization
- Permission mapping
- Record streaming for indexing
- Attachment ingestion
- Automatic RecordGroup creation

---

## Performance Improvements

Compared to the existing Zammad connector:

- Uses Zendesk Incremental Export API instead of timestamp checkpoints
- Sideloads users, groups and organizations to avoid N+1 API requests
- Prevents infinite incremental loops when cursors stop advancing
- Correct synchronization order:
  - Users
  - User Groups
  - Record Groups
  - Records

---

# 🐛 Bug Fixes

This PR also fixes several issues in the generated Zendesk DataSource.

### Fixed

- ✅ Query parameters were silently ignored (`query_params` → `query`)
- ✅ API errors now correctly return the response body (`response.text()`)
- ✅ Added pagination support to endpoints that previously returned only the first page
- ✅ Attachment downloads now validate the destination host before sending authentication headers
- ✅ Reindex now updates existing records instead of recreating them
- ✅ Prevent infinite cursor loops during incremental synchronization

---

# 🧪 Testing

## Unit Tests

Added **67 new unit tests**

| File | Tests |
|------|------:|
| `test_zendesk_client.py` | 25 |
| `test_zendesk_connector.py` | 42 |

### Regression Tests

Included tests for:

- Query parameter bug
- Attachment host validation
- Cursor loop protection
- Reindex behaviour
- Permission handling

---

## Full Test Suite

```text
23941 passed
2 skipped
0 failed
```

---

## Live Validation

Verified against a real Zendesk instance.

### Synced Data

| Entity | Count |
|---------|------:|
| Tickets | 22 |
| Articles | 27 |
| Organizations | 16 |
| Groups | 5 |
| Users | 46 |

---

### Validation Results

| Check | Result |
|--------|--------|
| Records Indexed | ✅ 49 / 49 |
| Failed Records | ✅ 0 |
| Qdrant Collection | ✅ Healthy |
| Record Groups | ✅ 31 |
| User Links | ✅ 46 |
| Permission Edges | ✅ Created |
| Orphan Records | ✅ 0 |
| SyncPoint | ✅ Persisted |

A second sync was also performed to verify incremental synchronization and cursor reuse.

---

# 📸 Screenshots

screenshot-2026-07-26_16-14-36

screenshot-2026-07-27_12-33-02

screenshot-2026-07-30_01-51-02
screenshot-2026-07-30_01-51-11
screenshot-2026-07-30_01-51-18
screenshot-2026-07-30_01-51-25

---

# 📊 Verification Queries

The following Cypher queries were used during validation.

<details>

<summary>Show Queries</summary>

```cypher
// No orphan records
MATCH (r:Record {connectorName:'ZENDESK'})
WHERE NOT (r)-[:BELONGS_TO]->(:RecordGroup)
RETURN r.recordName;

// Record group distribution
MATCH (rg:RecordGroup {connectorName:'ZENDESK'})
OPTIONAL MATCH (r:Record)-[:BELONGS_TO]->(rg)
RETURN rg.groupType, rg.groupName, count(r);

// Ticket metadata validation
MATCH (r:Record {connectorName:'ZENDESK'})-[:IS_OF_TYPE]->(t:Ticket)
RETURN t.status, count(*);
```

</details>

---

# 🚫 Out of Scope

This connector is **read-only**.

Not included in this PR:

- Ticket creation
- Ticket updates
- Comment creation
- Help Center editing
- OAuth authentication
- Webhooks
- Agent tools

These can be implemented in future PRs.

---

# ⚠️ Known Limitations

The following are platform limitations rather than connector issues.

- Ticket status, priority, tags and assignee are stored as graph metadata and are not currently used during semantic retrieval.
- Organization RecordGroups are created but tickets belong to only one RecordGroup, so organization-based navigation is limited.
- These limitations are already being addressed by the ongoing Knowledge Graph tooling work.

---

# 📁 Files Changed

| Area | Description |
|------|-------------|
| Connector | Added full Zendesk connector |
| DataSource | Fixed generated Zendesk client |
| Factory | Registered connector |
| Registry | Removed placeholder implementation |
| Constants | Added Zendesk constants |
| Value Mapper | Added Zendesk status mappings |
| Tests | Added 4 test files |

---

# ✅ Checklist

- [x] Full Sync
- [x] Incremental Sync
- [x] Ticket Sync
- [x] Help Center Sync
- [x] User Sync
- [x] Organization Sync
- [x] Permission Mapping
- [x] Record Groups
- [x] Search Indexing
- [x] Unit Tests
- [x] Integration Validation
- [x] Live Tested

## Summary by CodeRabbit

* **New Features**
  * Added Zendesk as a supported connector.
  * Enabled synchronization of users, groups, organizations, tickets, comments, attachments, and Help Center content.
  * Added API-token and OAuth authentication.
  * Added incremental synchronization, filtering, permissions, reindexing, and connection testing.
  * Added pagination and secure attachment downloads.

* **Bug Fixes**
  * Mapped the “hold” status to Blocked.
  * Improved reliability for paginated and incremental synchronization.


backend/python/app/connectors/sources/zendesk/connector.py
```diff
@@ -0,0 +1,1188 @@
...
+PAGE_SIZE = 100
+
+
+@ConnectorBuilder("Zendesk")\
+    .in_group(AppGroups.ZENDESK.value)\
+    .with_description("Sync tickets, comments, attachments, articles, users, and groups from Zendesk")\
+    .with_categories(["Help Desk", "Knowledge Base"])\
+    .with_scopes([ConnectorScope.PERSONAL.value, ConnectorScope.TEAM.value])\
+    .with_auth([
+        AuthBuilder.type(AuthType.API_TOKEN).fields([
```
darshangodase: Can you add oauth support also

backend/python/app/connectors/sources/zendesk/connector.py
```diff
@@ -0,0 +1,1188 @@
...
+                name="subdomain",
+                display_name="Subdomain",
+                placeholder="Enter your Subdomain",
+                description="The Subdomain from Zendesk instance",
+                field_type="TEXT",
+                max_length=2000,
+            ),
+        ])
+    ])\
+    .with_info(CONNECTOR_EMAIL_IDENTITY_INFO)\
```
darshangodase, line 173: verify does this connector email identity info required in zendesk

backend/python/app/connectors/sources/zendesk/connector.py
```diff
@@ -0,0 +1,1188 @@
...
+        .add_documentation_link(DocumentationLink(
+            "Zendesk API Token Setup",
+            "https://developer.zendesk.com/documentation/ticketing/introduction/authentication/",
+            "setup",
+        ))
+        .add_documentation_link(DocumentationLink(
+            "Pipeshub Documentation",
+            "https://docs.pipeshub.com/connectors/zendesk/zendesk",
+            "pipeshub",
+        ))
```
darshangodase, line 185: add documentation for this connector in pipeshub docs repo

backend/python/app/connectors/sources/zendesk/connector.py
```diff
@@ -0,0 +1,1188 @@
...
+        except Exception as e:
+            self.logger.error(f"Failed to initialize Zendesk connector: {e}", exc_info=True)
+            return False
+
+    async def _get_fresh_datasource(self) -> ZendeskDataSource:
+        if not self.external_client:
+            await self.init()
+        if not self.data_source:
+            raise Exception("Zendesk data source is not initialized")
+        return self.data_source
```
darshangodase, line 304: for api token does not required this method whenever you will add oauth support handle accordingly

backend/python/app/connectors/sources/zendesk/connector.py
```diff
@@ -0,0 +1,1188 @@
...
+
+    async def run_sync(self) -> None:
+        self.logger.info(f"Starting Zendesk sync for connector {self.connector_id}")
+        if not self.external_client:
+            await self.init()
```
darshangodase: if external client not available do not re init  throw exception

rish664:
@coderabbitai full review

backend/python/app/connectors/sources/zendesk/connector.py
```diff
@@ -0,0 +1,1188 @@
...
+SYNC_POINT_KEY = "zendesk_incremental"
+DEFAULT_INCREMENTAL_START_TIME = 1
+PAGE_SIZE = 100
+
+
+@ConnectorBuilder("Zendesk")\
+    .in_group(AppGroups.ZENDESK.value)\
+    .with_description("Sync tickets, comments, attachments, articles, users, and groups from Zendesk")\
+    .with_categories(["Help Desk", "Knowledge Base"])\
+    .with_scopes([ConnectorScope.PERSONAL.value, ConnectorScope.TEAM.value])\
```
darshangodase: Make Team scope only

backend/python/app/connectors/sources/zendesk/connector.py
```diff
@@ -0,0 +1,1188 @@
...
+
+        users, user_email_map = await self._fetch_users(datasource)
+        if users:
+            await self.data_entities_processor.on_new_app_users(users)
+        self.logger.info(f"Zendesk: synced {len(users)} users")
+
+        # Must run after on_new_app_users so the creator's USER_APP_RELATION edge is
+        # written alongside the Zendesk-sourced ones, and before any record is
+        # emitted so the group permission can be stamped onto them.
+        await self._ensure_creator_access()
```
darshangodase: This is for personal connectors only so remove it

backend/python/app/connectors/sources/zendesk/connector.py
```diff
@@ -0,0 +1,1188 @@
...
+
+        group_record_groups, group_user_groups = await self._fetch_groups(datasource, user_email_map)
+        if group_user_groups:
+            await self.data_entities_processor.on_new_user_groups(group_user_groups)
+        if group_record_groups:
+            await self.data_entities_processor.on_new_record_groups(group_record_groups)
+        self.logger.info(f"Zendesk: synced {len(group_record_groups)} groups")
+
+        org_record_groups, org_user_groups = await self._fetch_organizations(datasource)
```
darshangodase: If org record groups not using anywhere  then why creating it ?

backend/python/app/connectors/sources/zendesk/connector.py
```diff
@@ -0,0 +1,1188 @@
...
+        datasource = await self._get_fresh_datasource()
+
+        self.sync_filters, self.indexing_filters = await load_connector_filters(
+            self.config_service,
+            "zendesk",
+            self.connector_id,
+            self.logger,
+        )
+
+        users, user_email_map = await self._fetch_users(datasource)
```
darshangodase: do not pass datasource from here if you are adding oauth support then add wherever calling api's

backend/python/app/connectors/sources/zendesk/connector.py
```diff
@@ -0,0 +1,1188 @@
...
+        if users:
+            await self.data_entities_processor.on_new_app_users(users)
+        self.logger.info(f"Zendesk: synced {len(users)} users")
+
+        # Must run after on_new_app_users so the creator's USER_APP_RELATION edge is
+        # written alongside the Zendesk-sourced ones, and before any record is
+        # emitted so the group permission can be stamped onto them.
+        await self._ensure_creator_access()
+
+        group_record_groups, group_user_groups = await self._fetch_groups(datasource, user_email_map)
```
darshangodase: do not pass datasource from here if you are adding oauth support then add wherever calling api's

backend/python/app/connectors/sources/zendesk/connector.py
darshangodase: do not pass datasource from here if you are adding oauth support then add wherever calling api's

backend/python/app/connectors/sources/zendesk/connector.py
```diff
@@ -0,0 +1,1188 @@
...
+        self.logger.info(f"Zendesk: synced {len(group_record_groups)} groups")
+
+        org_record_groups, org_user_groups = await self._fetch_organizations(datasource)
+        if org_user_groups:
+            await self.data_entities_processor.on_new_user_groups(org_user_groups)
+        if org_record_groups:
+            await self.data_entities_processor.on_new_record_groups(org_record_groups)
+        self.logger.info(f"Zendesk: synced {len(org_record_groups)} organizations")
+
+        ticket_count = await self._sync_tickets(datasource)
```
darshangodase: do not pass datasource from here if you are adding oauth support then add wherever calling api's

backend/python/app/connectors/sources/zendesk/connector.py
```diff
@@ -0,0 +1,1188 @@
...
+        article_count = await self._sync_help_center_articles(datasource)
```
darshangodase: do not pass datasource from here if you are adding oauth support then add wherever calling api's

backend/python/app/connectors/sources/zendesk/connector.py
anupPradhan0: Added. Introduced a second AuthBuilder.type(AuthType.OAUTH).oauth(...) configuration alongside the existing API Token flow.

Since Zendesk's OAuth endpoints are tenant-specific, the authorize/token URLs are collected as auth fields (same approach as the ServiceNow connector). The client-side OAuth implementation (ZendeskRESTClientViaOAuth and the auth_type == "OAUTH" branch in build_from_services) already existed, so this change mainly wires it into the connector configuration.

backend/python/app/connectors/sources/zendesk/connector.py
anupPradhan0: Verified — yes, it's required.

This connector resolves identity by email throughout. Record permissions are emitted as Permission(email=...) for the requester, and on_new_user_groups matches users via get_user_by_email. If a user's PipesHub email differs from their Zendesk email, they won't receive access.

This same constant is already used by 29 other connectors, including Zammad, so no change was needed.

backend/python/app/connectors/sources/zendesk/connector.py
anupPradhan0: I haven't done this yet since the documentation lives in a separate repository. I'll raise a separate PR there.

I'll make sure the permission model is documented, including ticket visibility (group/requester/organization), exclusion of internal (public=false) comments, and the handling of draft or restricted Help Center articles.

backend/python/app/connectors/sources/zendesk/connector.py
anupPradhan0: Updated accordingly.

_get_fresh_datasource() is now the single entry point for obtaining the datasource, which makes it the right place for OAuth token refreshes. I also removed the datasource parameter from the sync helpers (_fetch_users, _fetch_groups, _fetch_organizations, _sync_tickets, _sync_help_center_articles), so each method resolves the datasource through _get_fresh_datasource(). That ensures a refreshed OAuth token is picked up immediately instead of reusing a stale datasource for the entire sync.

backend/python/app/connectors/sources/zendesk/connector.py
anupPradhan0: Fixed.

_get_fresh_datasource() now raises a RuntimeError if the external client or datasource hasn't been initialized instead of attempting to re-initialize them.

I also removed the duplicate initialization check from run_sync(), since _get_fresh_datasource() now handles this consistently.

This is safe because ConnectorFactory.create_and_initialize() already returns None if init() fails, so reaching run_sync() without a properly initialized datasource indicates a genuine bug that should be surfaced rather than hidden.

backend/python/app/connectors/sources/zendesk/connector.py
anupPradhan0: Done. Updated the connector to use only .with_scopes([ConnectorScope.TEAM.value]) and removed ConnectorScope.PERSONAL.

backend/python/app/connectors/sources/zendesk/connector.py
anupPradhan0: Removed, along with the four _connector_group_permission usages it fed.

I verified that ensure_connector_group_permission is only used by the *_personal connectors. Access is now derived entirely from Zendesk's own ACLs.

One behavioral change worth noting is that a PipesHub admin without a matching Zendesk user will no longer see synchronized records. That's expected for a Team-scoped connector, but it's worth calling out in the PR.

backend/python/app/connectors/sources/zendesk/connector.py
anupPradhan0: You're right. Nothing ever assigned records to org_* RecordGroups—tickets use group_* and Help Center articles use section_*, so the organization RecordGroups remained permanently empty.

I've removed them. _fetch_organizations() now returns only user groups, and organization-based visibility is represented by Permission(GROUP, external_id=f"org_{id}") on the ticket using ticket_data["organization_id"]. This also makes Zendesk's shared-organization visibility work correctly.

backend/python/app/connectors/sources/zendesk/connector.py
anupPradhan0: Done. _fetch_users() now resolves the datasource internally via await self._get_fresh_datasource() instead of receiving it as a parameter.

It also now returns a completeness flag so that a truncated user export never reaches on_new_user_groups, which rebuilds memberships from the supplied dataset.

backend/python/app/connectors/sources/zendesk/connector.py
anupPradhan0: Done. _fetch_groups(user_email_map) now resolves the datasource internally through _get_fresh_datasource().

It also reports whether the entire export completed successfully. Since list_group_memberships uses offset pagination and Zendesk stops beyond 10,000 records, membership rebuilds are now skipped if the export is incomplete.

backend/python/app/connectors/sources/zendesk/connector.py
anupPradhan0: Done. _fetch_organizations() now resolves the datasource internally instead of accepting it as a parameter.

Its return type has also been narrowed to user groups only, following the removal of unused organization RecordGroups.

backend/python/app/connectors/sources/zendesk/connector.py
anupPradhan0: Done. _sync_tickets() now resolves the datasource internally through _get_fresh_datasource().

backend/python/app/connectors/sources/zendesk/connector.py
anupPradhan0: Done. Both _sync_help_center_articles() and _sync_help_center_sections() now resolve the datasource internally instead of receiving it from run_sync().

This also prepares the connector for OAuth. A datasource resolved once in run_sync() would become stale if the access token were refreshed during a long-running sync. Resolving it inside each method ensures refreshed credentials are picked up automatically, with _get_fresh_datasource() acting as the single refresh point.

backend/python/app/connectors/sources/zendesk/connector.py
```diff
@@ -0,0 +1,1188 @@
...
+        datasource = await self._get_fresh_datasource()
+        content_url = record.weburl
+        if not content_url:
+            raise ValueError("Zendesk attachment missing content URL")
+        if not self._is_safe_zendesk_asset_url(content_url):
+            raise ValueError(
+                f"Refusing to fetch Zendesk attachment from untrusted host: {urlparse(content_url).hostname}"
+            )
+        headers = datasource.http.headers.copy()
+        response = await datasource.http.execute(HTTPRequest(url=content_url, method="GET", headers=headers))
```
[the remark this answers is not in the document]
anupPradhan0: Fixed, and tightened further than suggested.

_is_safe_zendesk_asset_url now requires https and rejects non-Zendesk hosts. I also added _is_tenant_api_url, which pins requests to the configured Zendesk subdomain. Previously, any *.zendesk.com host would have been accepted, so a URL pointing to another tenant's subdomain could have received this tenant's API credentials.

Authorization headers are now sent only to the configured API host. CDN downloads use headers = {} since those URLs are already pre-signed and don't require authentication.

I also changed FileRecord.weburl to no longer store the temporary content_url, since it's effectively a bearer token. Instead, _resolve_attachment_url() fetches a fresh signed URL via show_attachment whenever the file is downloaded.

Added coverage with test_cdn_download_withholds_api_credentials plus negative tests for foreign subdomains and plain HTTP URLs.

backend/python/tests/unit/connectors/sources/test_zendesk_connector.py
```diff
@@ -0,0 +1,1399 @@
...
+        datasource = MagicMock()
+        datasource.incremental_users = AsyncMock(return_value=_make_response(data={
+            "users": [{"id": 1, "email": "a@acme.com", "name": "A"}],
+            "after_cursor": "stuck",
+            "end_of_stream": False,
+        }))
+        users, _ = await zendesk_connector._fetch_users(datasource)
+        # First page consumed, second call detects the repeated cursor and stops.
+        assert datasource.incremental_users.await_count == 2
+        assert len(users) == 2
```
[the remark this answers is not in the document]
anupPradhan0: I haven't changed this because I don't believe the premise holds, but happy to revisit if I'm missing something.

On the first iteration the request is sent with cursor=None, and the server returns after_cursor="stuck". Fetching "stuck" on the next iteration is the first legitimate request for that page. Only after receiving "stuck" again does the connector detect that the cursor hasn't advanced (next_cursor == cursor) and terminate. These are two distinct requests; the duplicate records only appear because the mock returns the same payload twice.

The actual issue here was that the connector was using the time-based incremental endpoints (/incremental/{tickets,users,organizations}.json), which don't return after_cursor at all. As a result, every incremental sync stopped after the first 1,000 records. I've fixed that by switching to the /cursor.json endpoints, with start_time used only for the initial request and the returned cursor used for subsequent pages.

I also added tests/unit/sources/external/test_zendesk_data_source.py to verify the cursor endpoints and the start_time → cursor transition. Those tests would have caught the original issue.

backend/python/tests/unit/connectors/sources/test_zendesk_connector.py
```diff
@@ -0,0 +1,1399 @@
...
+
+    async def test_stream_record_dispatches_file(self, zendesk_connector):
+        _ready(zendesk_connector)
+        record = MagicMock()
+        record.record_type = RecordType.FILE
+        zendesk_connector._process_file_for_streaming = AsyncMock(return_value=b"bytes")
+
+        response = await zendesk_connector.stream_record(record)
+
+        assert response.media_type == MimeTypes.BLOCKS.value
```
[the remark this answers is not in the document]
anupPradhan0: Fixed.

stream_record() now handles RecordType.FILE using create_stream_record_response(..., mime_type=record.mime_type, ...), matching the approach used by the Jira connector. Previously, attachment bytes were always returned as MimeTypes.BLOCKS, even though the correct MIME type had already been stored on the FileRecord.

I also updated the test (test_stream_record_serves_file_with_its_own_mime_type) to assert the correct MIME type (text/plain) instead of preserving the previous incorrect behavior.

backend/python/app/connectors/sources/zendesk/connector.py
```diff
@@ -0,0 +1,1407 @@
...
+
+        Not re-initialising on failure: the factory never returns an uninitialised
+        connector, so reaching here uninitialised is a bug. Also the single place an
+        OAuth refresh belongs, hence callers re-resolve rather than hold one.
+        """
+        if not self.external_client or not self.data_source:
+            raise RuntimeError("Zendesk data source is not initialized")
+        if await self._oauth_token_rotated():
+            await self.init()
+        return self.data_source
```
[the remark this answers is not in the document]
anupPradhan0: implemented

backend/python/app/connectors/sources/zendesk/connector.py
```diff
@@ -0,0 +1,1407 @@
...
+                break
+            if not response.data:
+                break
+            payload = response.data
+            users_data.extend(self._extract_list(payload, "users"))
+            next_cursor = payload.get("after_cursor") or payload.get("cursor")
+            # A repeated cursor means the export is not advancing.
+            if payload.get("end_of_stream", True) or not next_cursor or next_cursor == cursor:
+                break
+            cursor = next_cursor
```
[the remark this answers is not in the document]
anupPradhan0: implemented

backend/python/tests/unit/sources/client/test_zendesk_client.py
```diff
@@ -0,0 +1,271 @@
...
+
+    def test_to_json(self):
+        resp = ZendeskResponse(success=True, data={"k": "v"})
+        assert json.loads(resp.to_json())["success"] is True
+
+    def test_defaults_are_none(self):
+        resp = ZendeskResponse(success=True)
+        assert resp.data is None
+        assert resp.error is None
+        assert resp.message is None
```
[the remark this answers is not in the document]
anupPradhan0: implemented

backend/python/app/connectors/sources/zendesk/connector.py
```diff
@@ -0,0 +1,1407 @@
...
+
+    async def _get_start_time(self) -> int:
+        sync_point = await self.records_sync_point.read_sync_point(SYNC_POINT_KEY)
+        start_time = sync_point.get("lastEndTime") or DEFAULT_INCREMENTAL_START_TIME
+        modified_filter = self.sync_filters.get(SyncFilterKey.MODIFIED) if self.sync_filters else None
+        if modified_filter:
+            value = modified_filter.get_value(default=None)
+            if isinstance(value, tuple) and value[0]:
+                start_time = max(int(start_time), int(value[0] / 1000))
+        return int(start_time)
```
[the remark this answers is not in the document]
anupPradhan0: SyncPoint.read_sync_point is annotated -> Dict[str, Any] and returns {} when no sync point exists (app/connectors/core/base/sync_point/sync_point.py:113); it never returns None. The suggested test would assert a state that can't occur. I've resolved the inconsistency in the other direction — dropped the redundant or {} in run_sync so both call sites rely on the declared contract.

PR #2850, opened 2026-07-31
fix(mail): bound SMTP and mail-backend timeouts to fail fast

anupPradhan0:
## Problem

`TestBulkInvite::test_bulk_invite_new_users_returns_200` fails in CI on both graph
backends after exactly 60s — the API never responds at all:

```
requests.exceptions.ReadTimeout: HTTPConnectionPool(host='localhost', port=3000):
Read timed out. (read timeout=60)
```

The test sends only 2 emails, so this is not a volume problem.

## Root cause

Two unbounded waits stack up on the invite path:

1. `mail.service.ts` calls the mail backend via axios with **no `timeout`** —
   axios defaults to waiting forever.
2. `mail.controller.ts` builds the nodemailer transport with **no timeouts**,
   inheriting nodemailer's **120s** connection timeout and a separate **30s**
   DNS timeout.

When the configured SMTP host is slow or unreachable, the TCP connect hangs
rather than being refused. 120s > the client's 60s read timeout, so the request
can never return in time — it holds the HTTP connection open and ties up a
request slot.

## Changes

`mail.controller.ts` — bound every stage of the SMTP handshake:

| Option | Value | Was |
|---|---|---|
| `dnsTimeout` | 10s | 30s (default) |
| `connectionTimeout` | 10s | 120s (default) |
| `greetingTimeout` | 10s | 30s (default) |
| `socketTimeout` | 20s | 10min (default) |

`mail.service.ts` — `timeout: 30_000` on the axios call to the mail backend.
Deliberately set *above* the SMTP timeouts so the inner layer fails first and
returns a real error instead of a generic axios timeout.

Successful sends are unaffected; this only bounds the failure path. A stalled
send now surfaces as a fast, reported failure instead of an open connection.

## Verification

- `tsc` clean; Node unit suite passes (0 failing)
- Timeout behaviour proven against a non-routable host: fails in **10.0s** with
  `ETIMEDOUT` instead of hanging
- With these applied, the integration test no longer hangs for 60s — it returns
  promptly

## Note for reviewers

`test_bulk_invite_new_users_returns_200` **will still fail on this PR**. This
change fixes the *hang*, not the underlying inability to reach an SMTP server
from CI runners, which cannot open outbound SMTP.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved email delivery reliability with connection, DNS, greeting, socket, and request timeouts.
  * Prevented email operations from hanging indefinitely when mail services are slow or unavailable.
  * Ensured stalled or failed email deliveries close their connections promptly.
  * Ensured suspicious-login alerts do not interrupt authentication error handling when email delivery fails.
  * Improved handling and logging of failed security alert notifications without exposing mail errors in authentication responses.


backend/nodejs/apps/src/modules/mail/controller/mail.controller.ts
```diff
@@ -17,6 +17,12 @@ import nodemailer from 'nodemailer';
 import { inject, injectable } from 'inversify';
 import { Logger } from '../../../libs/services/logger.service';
 import { AppConfig } from '../../tokens_manager/config/config';
+
+const SMTP_DNS_TIMEOUT_MS = 10_000;
+const SMTP_CONNECTION_TIMEOUT_MS = 10_000;
+const SMTP_GREETING_TIMEOUT_MS = 10_000;
+const SMTP_SOCKET_TIMEOUT_MS = 20_000;
+
```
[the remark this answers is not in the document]
anupPradhan0: Fixed

backend/nodejs/apps/src/modules/auth/services/mail.service.ts
```diff
@@ -64,6 +66,7 @@ export class MailService {
           'Content-Type': 'application/json',
         },
         data,
+        timeout: SEND_MAIL_TIMEOUT_MS,
```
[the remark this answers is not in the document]
anupPradhan0: Fixed

backend/nodejs/apps/tests/modules/mail/controller/mail.timeouts.test.ts
```diff
@@ -0,0 +1,132 @@
...
+      clock.restore();
+    }
+  });
+
+  it('still sends credentials when they are configured', async () => {
+    await build().emailSender(body, smtp);
+    const opts = createTransportStub.firstCall.args[0];
+
+    expect(opts.auth).to.deep.equal({ user: 'user', pass: 'pass' });
+  });
```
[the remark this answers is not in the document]
anupPradhan0: Fixed.

backend/nodejs/apps/tests/modules/auth/controller/suspiciousLoginAlert.test.ts
```diff
@@ -0,0 +1,113 @@
...
+    try { await runLogin(); } catch (e) { surfaced = e as Error; }
+
+    expect(surfaced, 'a security error must still be raised').to.exist;
+    // Never replaced by a mail-plumbing error.
+    expect(surfaced!.message).to.not.contain('timeout of 30000ms');
+    expect(surfaced!.message).to.contain('Incorrect password');
+    // Recorded rather than swallowed, and never retried into a duplicate.
+    expect(mockLogger.error.called).to.be.true;
+    expect(mockMailService.sendMail.callCount).to.equal(1);
+  });
```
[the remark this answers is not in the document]
anupPradhan0: done

vanshg685 (approved):

backend/nodejs/apps/src/modules/mail/controller/mail.controller.ts
```diff
@@ -17,6 +17,13 @@ import nodemailer from 'nodemailer';
 import { inject, injectable } from 'inversify';
 import { Logger } from '../../../libs/services/logger.service';
 import { AppConfig } from '../../tokens_manager/config/config';
+
+const SMTP_DNS_TIMEOUT_MS = 10_000;
```
rish664, line 21: @anupPradhan0 can you explain all the timeout in details

anupPradhan0: @rish664 
**The original bug:** the integration test died with `ReadTimeout` after 60s. Two
waits in the chain were unbounded, so a stalled SMTP host kept the HTTP request
open longer than the caller was willing to wait.

---

### Layer 1 — HTTP call to the mail service (30s)

`user_management/services/mail.service.ts` and `auth/services/mail.service.ts`

```ts
const SEND_MAIL_TIMEOUT_MS = 30_000;   // axios `timeout`
```

Axios has **no default timeout** — `0` means wait forever. If the mail service
hung, the caller hung with it. 30s bounds that. Both services get the same value
so the two invite paths behave identically.

### Layer 2 — nodemailer per-phase timeouts

`mail/controller/mail.controller.ts`

```ts
const SMTP_DNS_TIMEOUT_MS        = 10_000;  // default 30s
const SMTP_CONNECTION_TIMEOUT_MS = 10_000;  // default 2 min
const SMTP_GREETING_TIMEOUT_MS   = 10_000;  // default 30s
const SMTP_SOCKET_TIMEOUT_MS     = 20_000;  // default 10 min
```

Each covers one phase of the SMTP conversation:

| Phase | What it bounds |
|---|---|
| `dnsTimeout` | resolving the SMTP hostname |
| `connectionTimeout` | opening the TCP connection |
| `greetingTimeout` | waiting for the server's `220` banner |
| `socketTimeout` | inactivity once the conversation starts |

Nodemailer's defaults total **13 minutes** worst case — far past any HTTP
timeout above it. These bring it to 50s.

### Layer 3 — end-to-end deadline (25s)

```ts
const SMTP_SEND_DEADLINE_MS = 25_000;
```

This is the important one, and it's why layer 2 alone isn't sufficient.

`socketTimeout` measures **inactivity**, not elapsed time — it's implemented via
`socket.setTimeout()`, so it resets on every byte received. A server trickling
one byte every 19s never trips it. The phase budgets also run in *sequence*, so
they add up rather than cap.

`sendWithDeadline()` races `sendMail()` against a hard timer and closes the
transport on the losing path so the socket isn't leaked:

```ts
await Promise.race([transporter.sendMail(message), deadline]);
```

### The resulting budget

```
25s  SMTP send deadline      (hard cap)
30s  axios timeout           (caller gives up after the send should have failed)
60s  integration test client
```

Each layer fires before the one above it, so a failure surfaces as a real error
at the layer that caused it instead of bubbling up as an opaque timeout. 25s was
chosen to sit under the 30s axios timeout with margin.

Regression tests are in `mail.timeouts.test.ts` — one asserting the four option
values are set and each stays under the caller's timeout, one stalling
`sendMail` to confirm the deadline fires and the transport is closed.

anupPradhan0: Heads up — this PR grew a bit past the original timeout fix. The extra changes all
came out of the review feedback on it:

**1. Hard send deadline** (`SMTP_SEND_DEADLINE_MS`)
Review flagged that the per-phase timeouts could still exceed the caller's HTTP
timeout. Correct — `socketTimeout` measures inactivity and resets per byte, so it
was effectively unbounded. Added `sendWithDeadline()` to cap the whole send.

**2. Suspicious-login alerts made best-effort** (`userAccount.controller.ts`)
Review asked what happens when the alert mail fails. Turned out the auth
`MailService` *throws* (unlike the user_management one, which returns a status
code), so a mail timeout replaced the security response with a plumbing error —
the account was already blocked, but the caller saw an SMTP error instead. Now
wrapped in try/catch and logged.

**3. Timeout added to `auth/services/mail.service.ts` too**
There are two separate `MailService` classes; only one had the axios timeout.
Same fix applied to both so the invite and auth paths behave the same.

**4. Test suites consolidated to 2 tests each**
Per review, trimmed the regression tests down to one happy path + one failure path
per file.

rish664: i feel timeouts are much aggresive, can you see on internet what should be the ideal value

rish664: also if there are 1000 users added in bulk there would be 1000 http requests, can you check if same session is getting used or not?

anupPradhan0: @rish664 Yeah, 10s is definitely more aggressive than the usual SMTP defaults. The only reason I kept it that low is because email sending is still happening synchronously inside the HTTP request. If we increase it to 30s, we'd also have to increase the HTTP/Axios timeout, which starts taking us back toward the original long-running request issue.

Can we keep this PR focused on the current fix and close it? I'll raise a separate PR to implement the async mail architecture. Once email sending moves to the background, we can use proper SMTP timeout values without affecting request latency.

anupPradhan0: The HTTP side is already reusing connections (keepAlive), so we're not opening a new TCP connection for every request. However, the SMTP side is not reusing sessions—createTransport() is called for every email, so 1000 users means 1000 separate SMTP connections and handshakes.

Using Nodemailer's pool: true would help with connection reuse, but since the emails are still sent sequentially in the request path, I think it makes more sense to address both together in the async mail PR.

PR #2910, opened 2026-08-05
fix(it): harden storage connector incremental sync tests

shekharkadyan:
## Summary
- Hardens storage connector `TC-INCR-001` tests (S3, GCS, Azure Blob, Azure Files): settle IDLE + stable baseline, wait for sync completion, assert new file names with retries.
- Uses shared helper from **#2913** (`storage_incremental`).

## Depends on
**Merge [#2913](https://github.com/pipeshub-ai/pipeshub-ai/pull/2913) first.**  
Integration Tests use `pull_request_target`, so the workflow YAML (and the helper/unit suite it runs) must be on `main` before IT runs for this PR see that wiring.

## Test plan
- [x] Merge #2913
- [x] Rebase this branch onto `main`
- [ ] Run Integration Tests on this PR (`it-parallel` optional)
- [ ] Confirm S3 / GCS / Azure Blob / Azure Files `test_tc_incr_001_*` pass

## Summary by CodeRabbit

* **Tests**
  * Improved incremental synchronization coverage for Azure Blob, Azure Files, Google Cloud Storage, and Amazon S3.
  * Tests now use unique generated files and stable baselines to verify that new records become available without losing existing data.
  * Standardized validation and synchronization checks for more consistent, reliable results across supported storage connectors.
  * Unit tests no longer attempt to initialize live OAuth credentials before the local service stack is available.


integration-tests/unit/test_storage_incremental.py
```diff
@@ -0,0 +1,136 @@
...
+    )
+
+    with pytest.raises(AssertionError, match="at least 2 new records"):
+        await assert_incremental_new_files(
+            graph,
+            "conn-1",
+            before_count=5,
+            after_count=6,
+            new_names=["a.csv", "b.csv"],
+        )
```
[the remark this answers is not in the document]
shekharkadyan: Done — reduced to two regression tests:

1. `test_settle_baseline_and_assert_new_files_happy_path` — IDLE settle + name/count assert
2. `test_sync_until_names_visible_retries_then_fails_short_delta` — retry when names missing, plus short-delta rejection

Both pass locally via `pytest unit/`.

.github/workflows/integration-tests.yml
```diff
@@ -317,6 +317,8 @@ jobs:
           # distributed and skips teardown. -n 0 runs in-process and restores it.
           WORKERS="$PYTEST_WORKERS"
           if [ "$WORKERS" = "1" ]; then WORKERS=0; fi
+          # Helper unit tests (no Docker/services) — always run before the live suite.
+          pytest unit/ -v --tb=short --junitxml=reports/neo4j-unit-results.xml
```
siddhantota: Since these are simple unit tests and don't need the other services to be up and running, you can take them to just after the Set up Python and install dependencies and before Neo4j: Start services.
These tests anyways aren't done against live apis so we can fail the workflow if these fail even before starting other services.

This applies to the same block added to the arango ITs run as well

shekharkadyan: Good call — moved.

`pytest unit/` now runs once in a dedicated **Run helper unit tests** step right after dependency install and before any Neo4j/Arango `docker compose up`. Removed it from both live IT run blocks so we fail fast and don’t run the helpers twice.

shekharkadyan:
Split per Siddhant’s `pull_request_target` note:

1. **[#2913](https://github.com/pipeshub-ai/pipeshub-ai/pull/2913)** — workflow early `pytest unit/` + helper + unit tests (**merge this first**)
2. **This PR** — connector INCR hardening only (4 test files)

Please merge #2913, then we rebase this branch onto `main` and re-run ITs here.

abhishek-pipeshub (approved):

Issue #2975, opened 2026-08-15
Asking "how many…" strips every citation from the answer, including the parts that came from documents


shekharkadyan:
## What happens

Ask how many of something there is, and the answer comes back with **no citations at all** — including for the parts that came straight out of your documents and could have been cited.

The answer itself is fine. It is accurate, detailed, and drawn from real records. It just arrives with nothing attached to it.

That matters because "no citations" is how anything downstream tells a grounded answer from an invented one. An answer with no sources gets treated as unverified — so a perfectly good answer is thrown away, and the person is told the documents don't seem to cover it.

## Reproducing it

Three documents in the knowledge base. Same instance, same user, same corpus — only the question changes:

| Question | Citations |
| --- | --- |
| "What is each document about?" | **23**, then **6** |
| "How many documents are in the knowledge base?" | **0**, **0** |
| "How many distinct documents are there **and what is each one about?**" | **0**, **0**, **0** |

Row 3 is the one worth staring at. It is row 1 with four words added on the front. Those four words take the answer from 23 citations to none — and the per-document summaries, which were citable a moment earlier, are still sitting in the response, now unattributed:

```
There are 3 distinct documents available:

| Document Name | Summary |
| **onboarding-checklist** | Outlines a five-day onboarding checklist for new engineers,
                             covering equipment setup, account access, security training… |
| **harry potter 50 pages** | Contains the opening chapters of…                            |
| **Tetra**                 | A technology startup developing power electronics for the EV
                             market, with a focus on conditions in India                   |
```

Every one of those summaries came from a retrieved chunk. None is cited.

## Why this is not just cosmetic

A count is a reasonable thing to ask a knowledge base. "How many contracts mention X", "how many tickets are open on Y" — these are ordinary questions, and today every one of them produces an answer that anything checking for sources has to reject.

We hit this building a CLI that exits non-zero on an uncited answer, specifically so an agent cannot repeat unsourced prose as fact. That rule is doing its job here; the answer really is unverifiable as delivered. But the cause is upstream, and the result is that a correct answer gets reported to the user as "the documents do not appear to contain this."

## What I ruled out

**Not a parsing problem.** The raw answer text contains no `[source](refN)` markers and no `ref` substring at all. The model is not emitting citations that get lost later — it never emits them. (`app/utils/citations.py` handles multi-ref links, `[refN]`, and bare refs perfectly well when they do appear.)

[...]

PR #2976, opened 2026-08-15
fix(agents): stop one uncitable claim stripping citations from a whole answer

shekharkadyan:
Fixes the behaviour reported in #2975. **Partial** — measured, and honest about the remainder.

## The problem

Ask a knowledge base how many of something it has, and the answer arrives with no sources on anything — including the parts that came straight out of your documents.

Same corpus, same user, three documents. Only the question changes:

| Question | Citations |
| --- | --- |
| "What is each document about?" | **23**, then **6** |
| "How many documents are in the knowledge base?" | **0**, **0** |
| "How many are there **and what is each one about?**" | **0**, **0**, **0** |

Row 3 is row 1 with four words added. Those four words cost every citation, while the per-document summaries stay in the answer, now unattributed.

## Why

A count is not stated in any block — it comes from the shape of the result set. So it genuinely has no Citation ID, and the rule said:

> Omit the citation when no Citation ID is available for a fact.

Correct per-fact. But the model applies it to the **whole response**: one uncitable claim, and citations vanish from everything around it.

## The change

Scope the rule to the fact it is about, say what to do with a count, and require a citation on each item in a list or table.

## Measured

Same corpus, same question, same instance. Cited runs on "how many distinct documents are there and what is each one about?":

| | cited runs |
| --- | --- |
| stock | **0 / 14** |
| with this change | **6 / 16** (38%) |

Fisher one-sided **p = 0.0135**.

Variance is high — the same text produced 5/8 on one sample and 1/8 on the next, which is why both are pooled here rather than quoting the better one. Refusals were checked separately and stay correctly uncited (5/5 runs), so the rule does not push the model to invent a `refN` when the corpus genuinely has nothing.

**This does not fully fix it.** Three runs in five still strip every citation, and nothing downstream can tell those from a fabrication — #2975 should stay open, and the `exit 6` rule in the CLI should stay as it is.

Wordings that did **not** work, so nobody repeats them: scoping the rule without an imperative (0/8), scoping the imperative to "any part came from the context" (1/8), and appending an explicit refusal carve-out to the working text (0/8). The blunt "an answer must never end up with zero citations because one claim could not be cited" is the part carrying the effect.

## A wrong turn worth recording

My first attempt patched `app/modules/qna/prompt_templates.py:59`, which carries an identical instruction. It changed nothing across 8 runs, because that template is the non-agent path — `/chat/stream` runs through the agent loop, whose citation rules live here in `prompt_builder.py`. Two copies of the same rule in two prompts, one live for this route.

Worth knowing before someone else spends an afternoon on the other file. It may also be worth having one of them import the other, so a fix cannot land in the copy that is not running.

## Why it matters beyond the citation list

Anything that gates on citations has to reject an uncited answer, because nothing downstream can tell a stripped one from an invented one. The result is that a correct, complete answer gets reported to the user as "the documents do not appear to contain this."

Happy to run any alternative against this corpus — the reproduction is three questions and takes about a minute.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved citation handling for factual responses.
  * Preserved citations for citable facts while allowing uncited individual facts when no Citation ID is available.
  * Added citation requirements for every record-describing list item, table row, and bullet.

### DIFF
--- a/backend/python/app/agents/agent_loop/prompt_builder.py
+++ b/backend/python/app/agents/agent_loop/prompt_builder.py
@@ -197,7 +197,13 @@ used. Web search results use a different marker instead: `url/Citation ID: https
   carrying the real link.
 - Cite the block each fact actually came from; use a distinct ID for each distinct claim.
 - One ID per link, inline right after the key claim it supports. The system numbers them.
-  Omit the citation when no Citation ID is available for a fact.
+  Omit the citation for a fact that has no Citation ID — that fact only. Keep citing every
+  other fact that does have one. Counts and totals are derived
+  from the result set and have no Citation ID of their own: state the count uncited, then
+  cite each item you list under it. An answer must never end up with zero citations
+  because one claim could not be cited.
+- Every list item, table row, or bullet describing a record carries that record's citation,
+  e.g. `| RecordName | One-line summary. [source](refN) |`.
 - Keep prose readable: cite the specific claim it backs, not every clause or sentence in a
   row — a paragraph built from one source needs one citation at its end, not one per sentence.
   Never stack multiple `[source](...)` links back to back; if several blocks support the same


shekharkadyan:
Two of the three points were right and are fixed. The third I tested, and the suggestion measurably removes the fix — details below, because the numbers also force me to correct my own headline claim.

## Blocker — you were right, and it was my mistake

`docker-compose.override.yml` was in the PR. `git add -A` swept up an untracked lab file, and `git diff --stat` did not show it because it was untracked, so I committed it without seeing it. Dropped, and added to the lab clone's `.git/info/exclude` so it cannot happen again. The PR is one file now.

Landing that would have retargeted anyone running compose in that directory. Good catch.

## Lab document name — fixed

The worked example is now `| RecordName | One-line summary. [source](refN) |`, and the parenthetical "there are 3 documents" is gone too.

## The refusal exception — tested, and it costs the whole fix

Your reasoning was sound: Gate 5's correct negative is zero citations on purpose, and "never zero citations" could push the model to invent a `refN` on "I could not find it."

I ran your wording. It does not regress refusals. It also does not fix anything:

| Wording | cited runs |
| --- | --- |
| stock | **0 / 14** |
| mine — "must never end up with zero citations" | **6 / 16** |
| **yours** — cite what you can, "say so and cite nothing" | **0 / 8** |
| imperative scoped to "any part came from the context" | 1 / 8 |
| mine + your carve-out appended | 0 / 8 |

The blunt imperative is the part doing the work. Every softening I tried, including yours, returns it to stock behaviour.

So I tested the risk directly instead. Five refusal runs on the wording that works:

```
cites=0 | The provided documents do not contain information ab…
cites=0 | The provided documents do not contain information ab…
cites=0 | The provided documents do not contain information ab…
cites=0 | The provided context does not contain information ab…
cites=0 | The provided documents do not contain information ab…
```

The predicted regression does not occur. I have kept your carve-out out of the PR because it measurably costs the fix, but I would rather be wrong here than lose a correct negative — if you can produce a refusal that gains a citation under this wording, I will take the safer text and accept 0/8.

## Correcting my own claim

The PR said **0/6 → 5/8**. That does not replicate. The identical text re-run gave **1/8**.

Pooled, with a proper stock baseline:

- stock **0/14**
- this wording **6/16** (38%)
- Fisher one-sided **p = 0.0135**

So the effect is real but the variance is large, and my original number was a lucky sample presented as a result. I have rewritten the PR body accordingly. Anyone reading "5/8" and expecting to see it would not have.

## Agreed, no argument

- Not enough to close #2975 — three-in-five runs still strip everything, and nothing downstream can tell those from a fabrication.
- Exit `6` in `@pipeshub-ai/mcp@2.3.2` stays.
- `qna_prompt_simple` keeps the old line; it is the no-tools / Ollama degradation path (`chat_helpers.py:3133-3176`) and I am not touching it here.
- Not unifying the two templates in this PR — you are right that it would make the measurement uninterpretable.
- Updating #2975's stale "documents do not appear to contain this" line to match 2.3.2's wording.

shekharkadyan:
**Recommend closing this without merging.** A controlled A/B says the change does not work and carries a side effect. I am the one who opened it, so to be explicit: the earlier 5/8 was a fluke I should not have shipped a PR on.

## Method

Five answer shapes × two users × 6 runs, stock vs this branch — 120 calls, same corpus, same instance, harness identical across both arms.

| Probe | | stock cited/6 | patched cited/6 |
| --- | --- | --- | --- |
| grounded fact | preset | 5 | 5 |
| **true absence** | preset | **0** | **0** |
| **aggregate (the target)** | preset | **0** | **0** |
| whole-document summary | preset | 6 | 6 |
| compare two records | preset | 6 | 6 |
| all five | user B (no access) | 0 | 0 |

## It does not fix the target case

`0/6` both arms. Pooling every run today gives 6/22 patched against 0/20 stock — Fisher p = 0.014 — but that pooled number is not trustworthy: three independent samples of the same text gave **5/8, 1/8, 0/6**. One sample carries the entire effect and it has not reproduced since. That is a fluke, not a 27% improvement.

## It reduces citations on answers that were fine

The side effect nobody predicted. Per-run citation counts on a normal grounded question:

```
stock   : 13 13 13  0 13  1
patched : 13  1  1  1 13  0
```

Same on the summary probe, milder: 18.5 → 15.5 average, and tighter.

The likely reason is the instruction itself — "an answer must never end up with zero citations" reads as a floor, and the model satisfies a floor with one citation. Trading thirteen citations for one on a well-grounded answer is a real loss of traceability, and it happens on the common path rather than the rare one.

n=6 per arm, so this is a signal rather than proof. But there is no upside to weigh it against.

## What it does not do

The predicted risk did not materialise: **no invented citations on a genuine refusal**, 0/12 across both tokens, and no leakage to the user without access. That concern was reasonable and it is now measured.

## What this leaves

#2975 stays open and unfixed. The diagnosis in it still holds and is the useful output of this exercise:

- the failure is specific to **aggregate/counting** questions — summary and compare, both multi-record synthesis, cite perfectly at 6/6
- `list_files` is metadata-only and mints no refs, which is correct
- the same citation rule exists in two prompts and only one is live for chat, which is still worth collapsing

No prompt wording I tried moves this reliably, and I tried four. It needs a code-level fix, not a better sentence.

PR #3001, opened 2026-08-19
feat(mcp): log and echo X-Pipeshub-Request-Id header

### LINKED ISSUE #2945: Log the client-supplied X-Pipeshub-Request-Id header on /mcp requests

Comment here to claim it. One assignee. Node.js only; you do not need Python services.

### Why this exists

Clients send `X-Pipeshub-Request-Id` (the CLI mints a UUID per call). `/mcp` never logs it. The error path in `handleMCPRequest` logs `method` and `userId` only, so a pilot report of "empty answer at 14:30" cannot be grepped to the request.

### Files

- `backend/nodejs/apps/src/modules/mcp/controller/mcp.controller.ts`
- `backend/nodejs/apps/tests/modules/mcp/` (add or extend a test)

### What to change

1. Read `req.headers['x-pipeshub-request-id']` (header names are lower-cased in Express).
2. Log it on the existing MCP error log, and on a successful request at info or debug — enough that one grep finds the call.
3. Optionally echo the same value on the response (`X-Pipeshub-Request-Id`) so clients can confirm it landed.
4. Do not invent a second request-id if `x-request-id` is already present; log the client header **in addition**, because that is what the CLI prints in the transcript.

### How to verify

Unit-test the controller (or the route) with the header set and assert the logger (or response) includes it. `npm test` for the MCP tests.

### Out of scope

Do not change auth. Do not add a new tracing product.

prajyotdev530:
## Description

The `pipeshub` CLI sends an `X-Pipeshub-Request-Id` header on every /mcp
request, but the backend never logged it, so client-side traces couldn't
be correlated with the exact request that served them.

This PR reads the inbound `X-Pipeshub-Request-Id` header, logs it alongside
the existing method/userId metadata on both the success and error paths,
and echoes it back as a response header when the client sent one — so
clients can confirm the correlation landed. No new ID is generated when the
header is absent.

### Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)

## Related Issues
- Fixes #2945

## How Has This Been Tested?
Added unit tests covering: requestId included in success-path log, requestId
included in error-path log, response header echoed when present, response
header not set when absent.

### Test Configuration
- [x] Unit tests pass

## Summary by CodeRabbit

* **New Features**
  * Added support for propagating request IDs through successful MCP responses.
  * Request IDs are now included in related success and error logs for improved request tracking.

* **Bug Fixes**
  * Improved handling when optional authentication values are missing.
  * Enhanced error reporting with consistent error details and request context.
  * Ensured successful request processing completes before success status logging.

### DIFF
--- a/backend/nodejs/apps/src/modules/mcp/controller/mcp.controller.ts
+++ b/backend/nodejs/apps/src/modules/mcp/controller/mcp.controller.ts
@@ -72,16 +72,16 @@ export const handleMCPRequest =
         res.setHeader('X-Pipeshub-Request-Id', requestId);
       }
 
+      await transport.handleRequest(req, res, req.body);
+
       // Log success with request ID
       logger.info('MCP request completed', {
         method: req.method,
         userId: req.user?.userId as string | undefined,
         requestId,
       });
-
-      await transport.handleRequest(req, res, req.body);
     } catch (error: unknown) {
-      const err = error as Error;
+      const err = error instanceof Error ? error : new Error(String(error));
       const requestIdHeader = req.headers['x-pipeshub-request-id'] as
         | string
         | undefined;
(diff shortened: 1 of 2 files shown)


abhishek-pipeshub:
@prajyotdev530 can you also fix review comments

prajyotdev530:
i went through the coderabbit review comments and addressed the requested changes. pushed the updates to the same branch.

commits:
b8b314e — initial implementation
e840009 — addressed review comments
b69c9f5 — merged latest main

PR #3003, opened 2026-08-19
fix: add semantic:read and conversation:read to default MCP scopes

### LINKED ISSUE #2946: Stock MCP_SCOPES omits semantic:read and conversation:read, so they cannot be minted

Comment here to claim it. One assignee. Config / env only. Related: #2943 (search POST is still gated by `semantic:write` — do not mix that into this PR).

### Why this exists

`semantic:read` and `conversation:read` exist as route scopes but are missing from stock `MCP_SCOPES` / `DefaultMcpScopes`. A PAT or OAuth app cannot mint them on a default install, so a read-only client cannot request the names the routes use.

### Files

- `backend/nodejs/apps/src/modules/oauth_provider/config/scopes.config.ts` (`DefaultMcpScopes`)
- `backend/env.template` and `deployment/docker-compose/env.template` (`MCP_SCOPES=`)
- Compose defaults that hardcode the same comma-separated list (search `MCP_SCOPES=${MCP_SCOPES:-`)
- Existing tests in `backend/nodejs/apps/tests/modules/oauth_provider/` or tokens_manager

### What to change

Add `semantic:read` and `conversation:read` to the default mintable list everywhere that list is copied. Keep existing write scopes. Add or extend a test that the defaults include both read scopes.

### How to verify

`rg "MCP_SCOPES=" backend/env.template deployment/docker-compose` — both read scopes present. Node unit tests for default MCP scopes pass.

### Out of scope

Do not change `es.routes.ts` guards (that is #2943). Do not remove `semantic:write`.

Soban-2004:
Fixes #2946

## Problem

`semantic:read` and `conversation:read` are valid OAuth scopes but were missing from the default `MCP_SCOPES` allow-list.

As a result, requesting these scopes for a Personal Access Token could cause them to be silently dropped when they were not included in the configured default scope set.

## Fix

Added `semantic:read` and `conversation:read` to all 13 relevant copies of the default MCP scope list:

- `DefaultMcpScopes` in `backend/nodejs/apps/src/modules/oauth_provider/config/scopes.config.ts`
- `backend/env.template`
- `deployment/docker-compose/env.template`
- Relevant Docker Compose configuration files under `deployment/docker-compose/`

The new scopes are positioned alongside their corresponding write/chat scopes to keep the ordering consistent.

The `integrations/omnigent` scope list was intentionally left unchanged because it is a separate, narrower client-specific scope configuration.

## Testing

- Node.js/Mocha unit suite: 9254 passing, 0 failing
- TypeScript typecheck (`tsc --noEmit`): clean
- Verified the old default scope string has no remaining occurrences
- Verified the updated default scope list across the affected configuration files

## Summary by CodeRabbit

* **New Features**
  * Added read access for semantic search through the default MCP permissions.
  * Added conversation read access where supported.
  * Updated environment templates and deployment configurations to enable these permissions by default.


shekharkadyan:
@Soban-2004 can you resolve merge conflicts?

PR #3004, opened 2026-08-19
Fix/issue 2943 semantic read scope

### LINKED ISSUE #2943: semantic:write is required to run a read-only search, which the name contradicts
## Contributor task (`help wanted`)

Comment here to claim it. One assignee. Isolated Node.js change; not a first contribution.

### Why this exists

`POST /` on the semantic search router is a **read**. It is gated by `semantic:write`. `GET /` (history) correctly uses `semantic:read`. Stock `DefaultMcpScopes` includes `semantic:write` and omits `semantic:read`, so a read-only client cannot search without a write-named scope.

### Files

- `backend/nodejs/apps/src/modules/enterprise_search/routes/es.routes.ts` (around the `router.post('/', … requireScopes(SEMANTIC_WRITE)` for search)
- Tests next to that router
- `backend/nodejs/apps/src/modules/oauth_provider/config/scopes.config.ts` — only if `semantic:read` must be in `DefaultMcpScopes` for MCP clients to mint it (coordinate with #2946 if that is still open)

### What to change

Guard the search POST with `semantic:read` (or `semantic:read` **or** `semantic:write` if you must not break existing write-only tokens). Add a test that a token with only `semantic:read` can search, and that a token with neither cannot.

### Out of scope

Do not redesign the whole scope matrix. Do not change Python.

### Original report

`es.routes.ts` POST `/` used `semantic:write` while GET `/` used `semantic:read`.

prajyotdev530:
## Description

Fixes the `semantic:write` scope being required for the read-only search
endpoint (`POST /` in `es.routes.ts`). It now correctly requires
`semantic:read`, consistent with other read routes. Also added
`semantic:read` to the default `MCP_SCOPES` list in `env.template`, since
it was missing there.

## Related Issues

- Fixes #2943

## How Has This Been Tested?

Manually verified the search route now requires `semantic:read`, and that
`semantic:read` mints correctly via the default `MCP_SCOPES`. No other
routes or scopes were touched.

## Breaking Changes

Clients currently requesting `semantic:write` solely to perform search will
need to switch to `semantic:read`.

## Summary by CodeRabbit

* **New Features**
  * Updated semantic search access to use the `semantic:read` permission across supported integrations and API requests.
  * Added request ID tracking for MCP requests, including response headers and operational logs.

* **Bug Fixes**
  * Improved MCP error handling and request traceability when request IDs are provided.

* **Documentation**
  * Updated OAuth scope examples and integration guidance to reflect the revised permission.

### DIFF
--- a/backend/env.template
+++ b/backend/env.template
@@ -79,7 +79,7 @@ REFRESH_TOKEN_EXPIRY=720h
 
 
 # MCP OAuth scopes (comma-separated)
-MCP_SCOPES=openid,profile,email,offline_access,semantic:write,conversation:write,conversation:chat,kb:read,team:read,user:read,config:read,agent:read,agent:execute,connector:read
+MCP_SCOPES=openid,profile,email,offline_access,semantic:read,semantic:write,conversation:write,conversation:chat,kb:read,team:read,user:read,config:read,agent:read,agent:execute,connector:read
 
 # When using Kafka, enable the kafka profile to start Kafka + Zookeeper containers:
 #   COMPOSE_PROFILES=kafka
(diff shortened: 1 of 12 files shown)


prajyotdev530:
i will review these suggestions and will open a pull request after.

PR #3099, opened 2026-08-27
fix(auth): stop returning the OAuth client secret when JIT provisioning is disabled

shekharkadyan:
## What's wrong

`initAuth` is the endpoint the sign-in page calls before anyone has logged in. It tells the browser which sign-in methods exist and gives it the public details it needs to render them — so whatever it returns is readable by anyone who can reach the login page.

For OAuth, it was returning the client secret.

## Why it happens

The code does two things in the wrong order. First it puts the **whole** config into the response, then it removes the sensitive fields — but only inside an `if` that checks something unrelated:

```ts
// userAccount.controller.ts, before this change
authProviders[...] = configData;                    // ← full config, every time

if (configData?.enableJit === true) {               // ← unrelated condition
  jitEnabledMethods.push(method);
  jitConfig[mapping.key] = true;

  if (method === AuthMethodType.OAUTH) {
    const { clientSecret, tokenEndpoint, userInfoEndpoint, ...publicConfig } = configData;
    authProviders.oauth = publicConfig;             // ← the fix, only under JIT
  }
}
```

`enableJit` decides whether somebody signing in for the first time gets an account created automatically. That is a question about **provisioning**. Whether a field is safe to send to a browser is a question about **secrecy**. They have nothing to do with each other, so using one to gate the other means the cleanup simply does not run half the time.

When JIT is on, the secret is removed and everything is fine. When JIT is off, the line never executes and `clientSecret`, `tokenEndpoint` and `userInfoEndpoint` go out with the response.

## Why that combination is the bad one

Turning JIT **off** is the more locked-down setting: it means "do not let people create their own accounts, I will invite them". So the configuration an administrator chooses to be *more* careful is exactly the one that leaks. That is also what our documentation recommends for identity providers shared with a wider audience, which is how this was found.

## The fix

Strip the OAuth credentials whenever the config is added, rather than as a side effect of the JIT check:

```ts
if (method === AuthMethodType.OAUTH) {
  const { clientSecret, tokenEndpoint, userInfoEndpoint, ...publicConfig } = configData;
  authProviders.oauth = publicConfig;
} else {
  authProviders[mapping.key === 'azuread' ? 'azuread' : mapping.key] = configData;
}

if (configData?.enableJit === true) {
  jitEnabledMethods.push(method);
  jitConfig[mapping.key] = true;
}
```

The JIT block now does only what its name says. Nothing else changes: `jitEnabled` and `jitConfig` behave exactly as before, and the sign-in page still receives `providerName`, `clientId`, `authorizationUrl`, `scope` and `redirectUri` — everything it actually uses.

This also brings the two code paths into agreement. The authenticated version of this logic at `userAccount.controller.ts:1557` already strips unconditionally; only the unauthenticated one was conditional.

## Only OAuth is affected

I checked every other provider rather than assuming:

| Provider | What is stored | Secret? |
|---|---|---|
| OAuth | clientId, **clientSecret**, endpoints | **yes** |
| Google | clientId | no |
| Microsoft | clientId, tenantId | no |
| Azure AD | clientId, tenantId | no |
| SAML | entryPoint, certificate, emailKey | no — the certificate is the identity provider's public signing certificate |

So no other branch needs changing.

## Tests

Four tests in `userAccount.controller.test.ts` covering both JIT states, because the whole defect was that the two behaved differently:

- secrets absent when `enableJit: true`
- secrets absent when `enableJit: false`  ← the case that was broken
- the public fields the login page needs are still present
- `jitEnabled` still reports correctly either way

I confirmed they catch the bug by reverting the source change and re-running:

```
✔ ...when enableJit is true
1) ...when enableJit is false
   AssertionError: expected { providerName: 'Okta', …(6) } to not have property 'clientSecret'

3 passing, 1 failing
```

With the fix in place the whole controller suite passes: **163 passing**.

## After merging

Any OAuth client secret configured on an instance that had JIT disabled should be treated as exposed and **rotated at the identity provider**, then re-entered in PipesHub. Registering only the exact redirect URI at the provider limits what a leaked secret can be used for, and is worth confirming while you are there.

https://claude.ai/code/session_01HpNmn9X69EiiBEDGYTmLWZ

## Summary by CodeRabbit

* **Bug Fixes**
  * OAuth authentication settings no longer expose server-side credentials or endpoints in initialization responses.
  * OAuth provider details remain available for sign-in while sensitive information stays protected.
  * JIT-enabled and JIT-disabled configurations now report their status correctly.

* **Tests**
  * Added coverage verifying credential removal and JIT status across OAuth configurations.

### DIFF
--- a/backend/nodejs/apps/src/modules/auth/controller/userAccount.controller.ts
+++ b/backend/nodejs/apps/src/modules/auth/controller/userAccount.controller.ts
@@ -351,17 +351,22 @@ export class UserAccountController {
 
             // Always add to allowed list if config exists
             allowedMethods.push(method);
-            authProviders[mapping.key === 'azuread' ? 'azuread' : mapping.key] = configData;
+
+            // This response is served to unauthenticated callers, so the OAuth
+            // config must be stripped of its server-side credentials before it
+            // goes out. That is independent of JIT, which only controls whether
+            // an account is provisioned on first sign-in.
+            if (method === AuthMethodType.OAUTH) {
+              const { clientSecret, tokenEndpoint, userInfoEndpoint, ...publicConfig } = configData;
+              authProviders.oauth = publicConfig;
+            } else {
+              authProviders[mapping.key === 'azuread' ? 'azuread' : mapping.key] = configData;
+            }
 
             // Only add to JIT lists if enableJit is strictly true
             if (configData?.enableJit === true) {
               jitEnabledMethods.push(method);
               jitConfig[mapping.key] = true;
-
-              if (method === AuthMethodType.OAUTH) {
-                const { clientSecret, tokenEndpoint, userInfoEndpoint, ...publicConfig } = configData;
-                authProviders.oauth = publicConfig;
-              }
             }
           } catch (e) {
             this.logger.debug(`${method} config fetch failed, not adding to allowed methods`);
--- a/backend/nodejs/apps/tests/modules/auth/controller/userAccount.controller.test.ts
+++ b/backend/nodejs/apps/tests/modules/auth/controller/userAccount.controller.test.ts
@@ -1165,6 +1165,75 @@ describe('UserAccountController', () => {
     });
   });
 
+  describe('initAuth - OAuth credentials are never returned', () => {
+    const oauthConfig = (enableJit: boolean) => ({
+      providerName: 'Okta',
+      clientId: 'public-client-id',
+      clientSecret: 'SUPER-SECRET',
+      authorizationUrl: 'https://idp.example.com/authorize',
+      tokenEndpoint: 'https://idp.example.com/token',
+      userInfoEndpoint: 'https://idp.example.com/userinfo',
+      enableJit,
+    });
+
+    const arrange = (enableJit: boolean) => {
+      sinon.stub(Org, 'findOne').resolves({ _id: 'o1', isDeleted: false } as any);
+      sinon.stub(OrgAuthConfig, 'findOne').resolves({
+        orgId: 'o1',
+        authSteps: [{ order: 1, allowedMethods: [{ type: 'oauth' }] }],
+      } as any);
+      mockConfigService.getConfig.resolves({ data: oauthConfig(enableJit) });
+      mockSessionService.createSession.resolves({
+        token: 'session-oauth-123',
+        userId: 'u1',
+        email: 'user@example.com',
+        authConfig: [{ order: 1, allowedMethods: [{ type: 'oauth' }] }],
+        currentStep: 0,
+      });
+    };
+
+    // initAuth is unauthenticated. The server-side credentials must not appear
+    // in its response whether or not JIT provisioning happens to be enabled.
+    [true, false].forEach((enableJit) => {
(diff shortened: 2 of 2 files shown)


shekharkadyan:
Closing the two gaps I left open in the description: an over-the-wire confirmation, and an audit for the same defect elsewhere.

## Confirmed on the wire, not just in a unit test

The unit tests inspect what the controller hands to `res.json`. To be sure the field actually survives serialisation and leaves the process, I mounted the real controller in a real Express app and issued an ordinary unauthenticated `POST /api/v1/userAccount/initAuth` over HTTP.

**Unfixed `main`, JIT disabled** — HTTP 200, raw response body:

```json
{"currentStep":0,"allowedMethods":["oauth"],"message":"Authentication initialized",
 "authProviders":{"oauth":{"providerName":"Okta","clientId":"public-client-id",
 "clientSecret":"SUPER-SECRET-VALUE","authorizationUrl":"https://idp.example.com/authorize",
 "tokenEndpoint":"https://idp.example.com/token",
 "userInfoEndpoint":"https://idp.example.com/userinfo","enableJit":false}},
 "jitEnabled":false}
```

**Unfixed `main`, JIT enabled** — same request, secret absent:

```json
{"authProviders":{"oauth":{"providerName":"Okta","clientId":"public-client-id",
 "authorizationUrl":"https://idp.example.com/authorize","enableJit":true}},"jitEnabled":true}
```

**With this PR applied**, both JIT states return the second shape. No secret, no token endpoint, no userinfo endpoint, and the fields the sign-in page needs are untouched.

The request carried no credentials — just `{"email":"..."}` — and `POST /initAuth` has only `ValidationMiddleware.validate(...)` on it in `userAccount.routes.ts:50-53`. `authMiddleware` is constructed on line 48 but never applied to that route.

## The defect class does not repeat

The shape of this bug is a security cleanup nested inside a condition that is about something else. I searched for others:

**Destructure-and-drop strips** — only two in the Node service, both on this config, and after this PR both are unconditional:

```
userAccount.controller.ts:362   (this PR)
userAccount.controller.ts:1557  (already correct — authenticated path)
```

**Masking helpers** — every call site is gated on a `hideSecrets` ternary, which is at least a secrecy-shaped condition rather than an unrelated one. `shouldHideSecrets()` reads `process.env.HIDE_SECRET_CONFIG === "true"`, so masking is off unless an operator opts in — but those routes sit behind `authMiddleware.authenticate` + `requireScopes(CONFIG_READ)` + `userAdminCheck`, so an admin reading back a secret they can already write is a design choice, not an exposure. Nothing to change.

One loose end worth a separate look, not a security issue: `maskGoogleAuthConfig`, `maskMicrosoftAuthConfig`, `maskOAuthConfig` and `maskGithubAuthConfig` are defined in `maskConfigSecrets.ts` and have **zero call sites**. Someone wrote maskers for the auth configs and they were never wired up. The paths they would have covered are admin-gated, so nothing leaks, but they are either dead code to delete or an intention that was dropped.

So the unauthenticated leak was this one, and only this one.

abhishek-pipeshub (approved):

PR #3103, opened 2026-08-28
fix: remove client header fallback for user/org identity in toolset routes

igennova:
## Summary

`_get_user_context()` in `toolsets.py` and `mcp_servers.py` (and the `caller_id` lookup in the OAuth callback handler) fell back to reading `X-User-Id` / `X-Organization-Id` request headers whenever the verified JWT payload on `request.state.user` was missing that field:

    user_id = user.get("userId") or request.headers.get("X-User-Id")
    org_id = user.get("orgId") or request.headers.get("X-Organization-Id")

Those headers are plain, unsigned strings any caller can set — there'sno verification step. A caller holding any valid-but-narrowly-scoped JWT (e.g. a token type that doesn't carry `userId`/`orgId` claims) could set these headers directly and be treated as an arbitrary user/org on toolset and MCP-server endpoints.

## Fix

- Identity is now sourced only from `request.state.user` (the verified
  JWT payload). The header fallback is removed entirely.
- `_get_user_context()` now requires **both** `user_id` and `org_id` to  be present (previously only `user_id` was enforced), closing a related gap where a token with a user but no org would pass through with `org_id=None`.

## Verified no legitimate dependency

Grepped `backend/nodejs` and `frontend` — nothing sends
`X-User-Id`/`X-Organization-Id` to these routes. The only references
were in this repo's own unit tests, which asserted the old fallback
behavior and have been updated to assert it's now rejected with 401.

## Testing

Screenshot 2026-08-29 at 2 54 17 AM

## Summary by CodeRabbit

- **Security Enhancements**
  - Requests now use authenticated session context exclusively; user-provided identity headers are ignored.
  - Requests missing authenticated user or organization information are rejected with an authorization error.
  - OAuth callback validation now relies only on the authenticated session identity.

- **Bug Fixes**
  - Prevented unauthorized access through spoofed identity headers.
  - Updated validation behavior to consistently require complete authentication context.


igennova:
@abhishek-pipeshub this is a security-relevant fix (auth/identity handling in toolset routes), so flagging for priority review.

abhishek-pipeshub (approved):

PR #3111, opened 2026-08-30
docs: make the playbook's connector boilerplate constructible

shekharkadyan:
## The problem

The connector template in Step 6.2 of `CONNECTOR_INTEGRATION_PLAYBOOK.md` cannot be instantiated. Copy it, fill in your service, and it raises a `TypeError` before any of your own code runs.

Two of its calls do not match the classes it builds on.

**`super().__init__` is missing three arguments.** `BaseConnector.__init__` (`connector_service.py:56`) takes eight; the template passes five, leaving out `connector_id`, `scope` and `created_by`.

**`SyncPoint` is given the wrong keyword.** `SyncPoint.__init__` (`sync_point.py:32`) takes `connector_id`. The template passes `connector_name`, so once the first problem is fixed construction fails on the next line instead.

The unit-test helper added in Step 8.1 has the same gap, because it was written against the template rather than against the base class.

## The change

The template's signature now takes `connector_id`, `scope` and `created_by` and forwards them, matching how the shipped connectors do it — see `sources/linear/connector.py:76`. The `SyncPoint` call uses `connector_id`. The test helper passes all seven arguments.

Step 8.2's "Expected Output" also still showed the banner printed by a `__main__` script that is no longer in the document. It now shows what `pytest` actually prints.

## Verification

Rather than reading the signatures by eye, I extracted them and compared:

```
BaseConnector expects : app, logger, data_entities_processor, data_store_provider,
                        config_service, connector_id, scope, created_by
template super() sends: app, logger, data_entities_processor, data_store_provider,
                        config_service, connector_id, scope, created_by
                        -> order and count match

SyncPoint expects     : connector_id, org_id, sync_data_point_type, data_store_provider
template passes       : connector_id, org_id, sync_data_point_type, data_store_provider
                        -> match
```

## Note

`connector_name` is still passed to `FileRecord` further down the playbook. That one is correct — `Record` genuinely has a `connector_name` field (`models/entities.py:224`). Only the `SyncPoint` call was wrong.

## Summary by CodeRabbit

* **New Features**
  * Connector setup now supports explicit connector identifiers, scopes, and creator information.
  * Synchronization tracking is associated with the configured connector ID.
  * Connector creation accepts a preconfigured data processor and additional setup options.

* **Tests**
  * Updated test setup to use the new connector configuration requirements.
  * Refreshed sample output to show automated test execution with three passing tests.

### DIFF
--- a/CONNECTOR_INTEGRATION_PLAYBOOK.md
+++ b/CONNECTOR_INTEGRATION_PLAYBOOK.md
@@ -679,15 +679,22 @@ class YourConnector(BaseConnector):
         logger: Logger,
         data_entities_processor: DataSourceEntitiesProcessor,
         data_store_provider: DataStoreProvider,
-        config_service: ConfigurationService
+        config_service: ConfigurationService,
+        connector_id: str,
+        scope: str,
+        created_by: str,
     ) -> None:
         super().__init__(
-            YourConnectorApp(),
+            YourConnectorApp(connector_id),
             logger,
             data_entities_processor,
             data_store_provider,
             config_service,
+            connector_id,
+            scope,
+            created_by,
         )
+        self.connector_id = connector_id
 
         # TODO: Initialize your external API client reference
         self.external_client: Optional[Any] = None
@@ -695,7 +702,7 @@ class YourConnector(BaseConnector):
         # Initialize sync point for delta sync
         # See: https://github.com/pipeshub-ai/pipeshub-ai/blob/main/backend/python/app/connectors/core/base/sync_point/sync_point.py
         self.sync_point = SyncPoint(
-            connector_name=Connectors.YOUR_CONNECTOR,
+            connector_id=connector_id,
             org_id=self.data_entities_processor.org_id,
             sync_data_point_type=SyncDataPointType.RECORDS,
             data_store_provider=self.data_store_provider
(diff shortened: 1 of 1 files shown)


shekharkadyan:
Both points are fixed.

**App subclass.** `App.__init__` (`interfaces/connector/apps.py:7`) takes `(app_name, app_group_name, connector_id)`. The template defined `__init__(self)` and called `super()` with `app_name=` and `app_group=` keywords, so passing `connector_id` into `YourConnectorApp` raised `TypeError`.

There is a second fault in the same two lines. `AppGroup` is not the enum — `apps.py:21` defines it as a container class holding a list of `App`s, so `AppGroup.YOUR_VENDOR` raised `AttributeError` in any case. The enum is `AppGroups` in `config/constants/arangodb.py`. Both now match `LinearApp` (`linear/common/apps.py:6`), and the unused `AppGroup` import is gone.

**Factory method.** `create_connector` accepted three arguments and built the entities processor itself. `connector_factory.py:229` builds and initialises that processor and passes it in, along with `connector_id`, `scope` and `created_by`, so a connector written to the template failed at creation time. The signature now matches, and takes `**kwargs` so later additions to the factory call do not break it.

**Verification.** The real signatures were extracted and compared against every step of the construction chain:

```
App super() arity + enum name      MATCH
YourConnector.__init__ vs base     MATCH
SyncPoint keywords                 MATCH
create_connector accepts all       MATCH
```

**Non-blocking.** `addopts = -q` means pytest prints progress dots, so the expected output no longer shows a file name.

`self.connector_id = connector_id` after `super()` stays. It is redundant, as noted, but `LinearConnector` does the same and a template is more useful when it matches the shipped connectors.

abhishek-pipeshub (approved):

PR #3178, opened 2026-09-05
test(connectors): hold every registered connector to the BaseConnector contract

shekharkadyan:
## In one line

Checks that all 50 connectors still fit the shape the code expects — no accounts, no passwords, about two seconds.

## Background

PipesHub connects to around 50 outside services — Google Drive, Slack, Jira and so on — to copy documents in so people can search across everything. Each connection is a *connector*.

Properly testing one means running it against the real service, which needs an account there. That is why only 16 of the 50 are tested: the rest would each need an account bought and maintained.

But some things about a connector can be checked without touching the service at all: that it is registered correctly, that it implements everything the shared interface requires, and that its setup function accepts the arguments it will actually be called with.

## What this adds

189 checks, needing no credentials.

| Check | The problem it catches |
|---|---|
| Each registered connector is a real connector class | A file renamed or moved, leaving the registration pointing at nothing usable |
| Nothing required is left unimplemented | A method added to the shared interface that most connectors never got |
| The setup function accepts the arguments it is given | A renamed parameter that only fails when someone configures that source |
| Registration names are safe to look up | A name with a capital or a space that quietly fails to match |

Plus two repository-wide rules: no two registration names point at the same class (almost always a copy-paste of an existing entry), and no connector talks to a database directly instead of going through the shared interface — a connector that does works on whichever setup the author had and fails on everyone else's.

## Why these matter

These are failures that do not show up when the code loads. They appear the first time somebody syncs that one source — on the connector nobody happened to try.

## What it found

**Seven registered connectors cannot sync at all.** `airtable`, `calendar`, `docs`, `forms`, `meet`, `slides` and `zendesk` are placeholders:

```python
class ZendeskConnector:
    def __init__(self) -> None:
        self.name = "Zendesk"

    def connect(self) -> bool:
        print(f"Connecting to {self.name}")
        return True
```

They are not inert. `connectors_main.py:229` adds them to the live registry at startup and `router.py:2659` serves them to the interface — so users are offered seven connectors that cannot work.

**This change does not alter that behaviour.** The seven are recorded in a list so these checks guard the other 43 rather than failing on a known issue. Two further checks keep that list honest: one fails if a listed name stops being registered, and one fails if a placeholder starts working, so the exemption is removed rather than left hiding a connector.

Whether to build them, hide them or remove them is a product decision.

## How to test it

```bash
cd backend/python
pytest tests/unit/connectors/test_connector_contract.py -v
```

Expect 189 passing in about two seconds.

To see them catch something, add a new required method to `BaseConnector` — every connector that does not implement it fails by name.

## Anything to worry about

No. This adds one test file. No product code changes.


shekharkadyan:
**Integration failure here is #3195, not this change.**

Six pull requests were approved for integration testing within a few minutes of each other. The object-store connector suites all sync from one shared, pre-provisioned bucket whose name is a constant (`RESOURCE_NAME` at `connector_lifecycle.py:75`), and the workflow's concurrency group is per-pull-request — so those runs uploaded, renamed and cleared each other's objects.

The signature is `BlobNotFound` on the rename and move cases, where the helper cannot find an object the same suite uploaded moments earlier:

```
azure.core.exceptions.ResourceNotFoundError: The specified blob does not exist.
ErrorCode:BlobNotFound
```

The same tests fail identically on pull requests that touch nothing related — #3177 only adds a CI step for agent evals, #3178 only adds unit tests — which is what rules this change out as the cause.

Every other check passes. Full evidence and two possible fixes are in #3195.

abhishek-pipeshub (approved):

PR #3203, opened 2026-09-07
test(connectors): add the GitHub Teams and GitLab suites the environment was built for

shekharkadyan:
## The test environment already exists

Pull request #3157, *"Github and Gitlab It environment setup"*, prepared everything these connectors need:

- **eleven secrets** exported to the integration job — token, org, three repositories for GitHub; token, instance URL, group, subgroup and two projects for GitLab
- both pytest markers registered in `pytest.ini`
- both offered as `workflow_dispatch` choices
- `env.local.template` documenting the required scopes and, in detail, the repository permissions the token owner needs

No test suite followed. Nothing in the repository reads any of those secrets, and both connectors sit on the uncovered list. This adds the suites they were provisioned for.

**Nothing outside `integration-tests/connectors/` changes** — the markers and dispatch choices were already in place.

## Scope: read-only, content-agnostic

The organisation, group, repositories and projects are shared fixtures these suites do not own. Nothing is created there, nothing is deleted, and no assertion names a particular issue or file — a test that hard-coded their contents would fail whenever somebody pushed, which is noise rather than a connector regression.

What remains is the failure mode that matters most for connectors like these: **they break when the provider changes**. A token expires, a scope is withdrawn, an API is retired, a members endpoint changes shape. None of that involves a change to this repository, and nothing else in CI would notice.

## What is covered

| Case | What it proves |
|---|---|
| TC-AUTH-001 | The token is accepted, its scopes still cover what the connector reads, and the provider's APIs still answer in the expected shape |
| TC-GRAPH-001 | Records are attached to groups with app edges and no orphans — an orphaned record indexes but is invisible to search |
| TC-PERM-001 | Permission edges exist, so access control synced and not only content |
| TC-REPOS-001 / TC-PROJECTS-001 | The configured repositories or projects appear in what was synced |

**TC-PERM-001 is the one worth having.** If permission sync stopped, content would still index and search would still return results, while access decisions were being made against nothing.

Its docstring points back at the setup notes from #3157, because the likeliest cause of a sudden drop to zero edges is the token losing the access those notes require, rather than a code change:

> **GitHub:** the token owner needs PUSH access on the repositories — the connector resolves permissions through `/collaborators`, and without push it silently falls back to the repository's visibility instead.
>
> **GitLab:** the token owner must be Owner or Maintainer on the group, because reading project members requires it.

## Two details that needed reading the clients

**The builders declare OAuth; the clients accept a token.** Both connectors declare `AuthType.OAUTH`, but that is only the default. Each client branches on `auth.authType` at runtime and supports a personal access token — which is what #3157 provisioned. The defaults differ, so this matters:

| | Default `authType` | |
|---|---|---|
| GitHub (`github.py:192`) | `API_TOKEN` | already correct |
| GitLab (`gitlab.py:284`) | `OAUTH` | must be set explicitly |

**Both clients require a `credentials` block even on the token path.** `github.py:189` and `gitlab.py:280` raise on an empty `credentials` *before* reaching the `API_TOKEN` branch — which then never reads it. So a token-configured connector has to pass a dict it does not use.

That looks like a defect rather than intent. The fixtures pass one rather than working around it in product code, since changing that from inside a test pull request would be the wrong place for it.

## Coverage

Two more connectors covered with nothing to procure. Combined with the other suites in flight, connector integration coverage moves from 16 of 50 to 26.

## What is not verified here

Neither suite can run without the accounts, so CI is their first real execution — as it is for any suite reading these secrets, since none has ever read them. Both skip cleanly when the credentials are absent, so forks and local runs are unaffected.

## Summary by CodeRabbit

- **Tests**
  - Added integration coverage for GitHub Teams and GitLab connectors.
  - Verified token authentication and successful synchronization.
  - Added checks for graph consistency, permissions, and relationships between synced records.
  - Confirmed configured repositories and projects are included in synchronization results.
  - Added test setup that skips runs when required credentials or configuration are unavailable and cleans up connector resources after execution.

### DIFF
--- a/integration-tests/connectors/github_teams/conftest.py
+++ b/integration-tests/connectors/github_teams/conftest.py
@@ -1,0 +1,145 @@
+# pyright: ignore-file
+
+"""GitHub Teams connector fixtures.
+
+The credentials this needs were provisioned in #3157, which also registered the
+``github_teams`` pytest marker and documented how the account must be set up.
+No suite followed, so nothing has been reading them.
+
+Read-only. The organisation and its repositories are shared fixtures that this
+suite does not own, so it creates nothing and deletes nothing there; only the
+connector and its graph data are cleaned up.
+"""
+
+import os
+import uuid
+from collections.abc import AsyncGenerator
+from typing import Any
+
+import pytest
+import pytest_asyncio
+from connector_lifecycle import destructor
+from helper.graph_provider import GraphProviderProtocol
+from helper.graph_provider_utils import wait_until_graph_condition
+from pipeshub_client import PipeshubClient  # type: ignore[import-not-found]
+
+# Set by #3157. env.local.template documents the required scopes and, more
+# importantly, that the token owner needs PUSH access on every repository: the
+# connector reads /collaborators to resolve permissions, and without push it
+# silently falls back to the repository's visibility instead.
+ENV_TOKEN = "GH_TEAMS_TEST_TOKEN"
+ENV_ORG = "GH_TEAMS_TEST_ORG"
+ENV_PRIMARY_REPO = "GH_TEAMS_TEST_PRIMARY_REPO"
+ENV_PUBLIC_REPO = "GH_TEAMS_TEST_PUBLIC_REPO"
+
+REQUIRED = (ENV_TOKEN, ENV_ORG, ENV_PRIMARY_REPO)
+
+
+@pytest.fixture(scope="session")
+def github_teams_settings() -> dict[str, str]:
+    missing = [name for name in REQUIRED if not os.getenv(name)]
+    if missing:
+        pytest.skip(f"GitHub Teams credentials not set: {', '.join(missing)}")
+    settings = {name: os.environ[name] for name in REQUIRED}
+    settings[ENV_PUBLIC_REPO] = os.getenv(ENV_PUBLIC_REPO, "")
+    return settings
+
+
+def expected_repos(settings: dict[str, str]) -> list[str]:
+    """Repositories the sync is expected to reach, from configuration."""
+    names = [settings[ENV_PRIMARY_REPO]]
+    if settings.get(ENV_PUBLIC_REPO):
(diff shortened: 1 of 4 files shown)


shekharkadyan:
Closing. This cannot pass, and the reason is in the product rather than the tests — filed as #3204.

Review found four blocking problems. All four are real; I verified each against the source before accepting them:

**1. `connector_type="GithubTeams"` 404s.** `ConnectorBuilder("GitHub Teams")` registers the name with a space, and `get_connector_metadata` does an exact membership test (`connector_registry.py:1246`) with no normalisation. I had checked the *factory* registry key, which is lowercased and space-stripped — but create never reaches the factory. GitLab's `"GitLab"` happened to be correct.

**2. `auth_type="API_TOKEN"` is rejected with 400.** Both builders declare only OAuth, and `router.py:3797` validates the requested type against that list. That the clients accept a token at runtime does not matter: the registry gate comes first. Linear works because it declares *both* auth types.

**3. The `credentials` workaround cannot work.** `_prepare_connector_config` hardcodes `"credentials": None` (`router.py:3583`), so the dict passed by a fixture is discarded. Both clients then raise on the empty value (`github.py:188`, `gitlab.py:279`) *before* reaching the `API_TOKEN` branch that never reads it. I had noticed that ordering and worked around it in the fixture; the workaround does not survive the API.

**4. The repository assertion looks in the wrong place.** Repository identity is a `RecordGroup.name` — `repo.full_name` at `github_teams/projects.py:410`, `project.path_with_namespace` at `gitlab/projects.py:509` — not a record name or path. A successful sync could fail that assertion, and `fetch_record_paths` is capped at 200 besides. Substring matching would also let `api` match `api-client`.

The right sequence is #3204 first — declare `API_TOKEN` on both builders, and move the `credentials` check into the OAuth branch where the value is used. The suites can then be reopened against `RecordGroup` names and `wait_for_sync_completion` instead of `count_records > 0`.

Reopening this before that would mean maintaining tests that cannot run.

PR #3232, opened 2026-09-10
Fix/issue 3225 public client revoke

### LINKED ISSUE #3225: No way to see or revoke a signed-in device or OAuth application
## What a person runs into

Someone signs in a coding agent with the new device flow — they open `/oauth/device`, type the code from their terminal, and click Allow. A week later they want to disconnect that agent, or they realise they approved a code somebody else sent them.

There is nowhere to do it. The dashboard has no list of connected devices or applications, and no way to revoke one. An administrator has no way to do it on the user's behalf either.

Personal access tokens do have this: you can list your own tokens and delete them, and an admin can revoke someone else's. Tokens issued through OAuth — including every device grant — have neither.

## Why the existing `/revoke` endpoint doesn't cover it

`POST /api/v1/oauth2/revoke` exists and implements RFC 7009 correctly, but it solves a different problem. It lets **a client throw away a token it is already holding**. It cannot help a person who wants to cut off a token held by someone else, because they would have to supply that token to revoke it, and they don't have it. There is also no way to enumerate what is currently active, so there is nothing to point it at.

## And a public client cannot call it at all

This part looks like a straightforward bug rather than a design gap.

`revoke` calls `verifyClientCredentials(client_id, client_secret)`, which unconditionally decrypts the stored secret and requires a timing-safe match. It never checks whether the app is a public client.

The first-party device client `pipeshub-agent` is created with `isConfidential: false`, and its generated secret is deliberately never handed out — the code says so: *"Schema still requires a stored secret; it is never used to authenticate."*

So an agent that signed in through the device flow has no `client_secret` and cannot revoke its own token. RFC 7009 §2.1 says a public client authenticates at the revocation endpoint the same way it does at the token endpoint, which for a public client is `client_id` alone. Today `/revoke` is unusable by exactly the client type the device flow was built for.

## Why it matters more than it looks

The device flow is the one grant a person can be talked into approving. They receive an eight-character code, they are already signed in, and the consent screen shows a trustworthy name. The anti-phishing copy added in #3197 is the right mitigation and it will stop most of it — but when it doesn't, there is currently no way to undo the result. The token also carries a refresh token, so it outlives the access-token window.

A control that can be socially engineered should be one you can reverse.

## Suggested scope

1. **Let public clients revoke their own tokens.** In `verifyClientCredentials`, or at the `revoke` call site, accept `client_id` alone when the app is not confidential. Smallest fix, restores RFC 7009 conformance, and gives agents a clean sign-out.
2. **Let a person see and end their own sessions.** A list of active OAuth grants for the signed-in user — application name, scopes, when it was approved, last used — with a revoke action. This is the piece a phishing victim actually needs.
3. **Let an administrator do the same for their organisation**, mirroring `adminRevokeToken` for PATs.

Item 1 is small and self-contained. Items 2 and 3 are the real work and could land separately.

<details>
<summary>Evidence</summary>

**PAT revocation exists** — `backend/nodejs/apps/src/modules/oauth_provider/routes/pat.routes.ts`: `GET /` lists the caller's tokens, `DELETE /:tokenId` revokes one, plus an admin revoke path. Backed by `revokeToken` and `adminRevokeToken` in `pat.service.ts`.

**OAuth routes** — `oauth.provider.routes.ts` exposes `/authorize`, `/token`, `/introspect`, `/revoke`, `/userinfo`. There is no list endpoint and no user- or admin-initiated revoke.

**`/revoke` requires a client secret** — `oauth.provider.controller.ts` `revoke()`:
```js
const { token, token_type_hint, client_id, client_secret } = req.body
await this.oauthAppService.verifyClientCredentials(client_id, client_secret)
await this.oauthTokenService.revokeToken(token, client_id, token_type_hint)
```

**`verifyClientCredentials` has no public-client branch** — `oauth.app.service.ts:416`:
```js
const storedSecret = this.encryptionService.decrypt(app.clientSecretEncrypted)
...
if (!crypto.timingSafeEqual(storedHash, providedHash)) {
  throw new InvalidClientError('Invalid client credentials')
}
```

**The first-party device client is public** — `oauth.first_party_device.service.ts` creates it with `isConfidential: false` and comments that the stored secret "is never used to authenticate".

Found while reviewing #3197 and the EE port. Neither PR introduces this; the device grant just makes it matter.
</details>

Saneajayy:
## Description

This PR fixes a bug where public clients (like the first-party device flow) were unable to revoke their own OAuth tokens because the `/api/v1/oauth2/revoke` endpoint unconditionally required and checked a `client_secret`. 

By definition, public clients do not authenticate with a secret. This PR updates `verifyClientCredentials` to elegantly handle public clients by skipping the secret check, which automatically fixes the `/revoke` and `/introspect` endpoints while keeping the logic strictly RFC 7009 compliant.

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update (Is there an addition/change in API Request, Response? If yes, documentation is required)
- [ ] Performance improvement
- [x] Code refactoring
- [x] Security fix

## Related Issues

- Fixes #3225

## How Has This Been Tested?

- Updated the existing mock configurations in `oauth.provider.controller.test.ts` to assert that `verifyClientCredentials` strictly enforces rejection for confidential clients when a secret is missing.
- Updated `mockApp` objects in `oauth.app.service.test.ts` to explicitly have `isConfidential: true` when testing cryptography/secret matching logic.
- Ran the full `npm run test` suite in `backend/nodejs/apps` which completed successfully with 100% passing tests.

### Test Configuration

- [x] Unit tests pass
- [ ] Integration tests pass
- [ ] Manual testing completed
- [ ] Cross-browser testing (if applicable)
- [ ] Mobile responsiveness tested (if applicable)

## Core Functionality Testing

- [x] OAuth Public clients can successfully hit the `/revoke` endpoint without providing a `client_secret`.
- [x] OAuth Confidential clients are still strictly required to provide a `client_secret`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Public OAuth clients can use authorization code, refresh token, and token revocation flows without providing a client secret.
  * Confidential clients must still provide valid client credentials; missing or invalid secrets return an authorization error.
  * Public clients cannot use the client credentials grant; requests are rejected with an authorization error.
* **Tests**
  * Expanded coverage for public and confidential client credential handling, including missing-secret scenarios and grant restrictions.

### DIFF
--- a/backend/nodejs/apps/src/modules/oauth_provider/controller/oauth.provider.controller.ts
+++ b/backend/nodejs/apps/src/modules/oauth_provider/controller/oauth.provider.controller.ts
@@ -399,13 +399,8 @@ export class OAuthProviderController {
       )
     }
 
-    // Confidential clients must provide client_secret
-    if (app.isConfidential) {
-      if (!clientSecret) {
-        throw new InvalidClientError('client_secret required for confidential clients')
-      }
-      await this.oauthAppService.verifyClientCredentials(clientId, clientSecret)
-    }
+    // Verify client credentials (public clients will skip secret check)
+    await this.oauthAppService.verifyClientCredentials(clientId, clientSecret)
 
     // Exchange code for tokens
     const codeResult = await this.authorizationCodeService.exchangeCode(
@@ -571,13 +566,8 @@ export class OAuthProviderController {
       )
     }
 
-    // RFC 6749 Section 6: Confidential clients MUST authenticate
-    if (app.isConfidential) {
-      if (!clientSecret) {
-        throw new InvalidClientError('client_secret required for confidential clients')
-      }
-      await this.oauthAppService.verifyClientCredentials(clientId, clientSecret)
-    }
+    // Verify client credentials (public clients will skip secret check)
+    await this.oauthAppService.verifyClientCredentials(clientId, clientSecret)
 
     // RFC 9700: For public clients, refresh tokens MUST be sender-constrained
     // or use rotation. We implement rotation in oauthTokenService.refreshTokens()
--- a/backend/nodejs/apps/src/modules/oauth_provider/controller/oauth.provider.controller.ts
+++ b/backend/nodejs/apps/src/modules/oauth_provider/controller/oauth.provider.controller.ts
@@ -474,6 +474,10 @@ export class OAuthProviderController {
       clientSecret,
     )
 
+    if (!app.isConfidential) {
+      throw new InvalidClientError('client_credentials grant requires a confidential client')
+    }
+
     // Verify grant type is allowed
     if (!this.oauthAppService.isGrantTypeAllowed(app, 'client_credentials')) {
       throw new UnsupportedGrantTypeError(
--- a/backend/nodejs/apps/src/modules/oauth_provider/services/oauth.app.service.ts
+++ b/backend/nodejs/apps/src/modules/oauth_provider/services/oauth.app.service.ts
@@ -415,9 +415,17 @@ export class OAuthAppService {
    */
   async verifyClientCredentials(
     clientId: string,
-    clientSecret: string,
+    clientSecret?: string,
   ): Promise<IOAuthApp> {
     const app = await this.getAppByClientId(clientId)
+
+    if (!app.isConfidential) {
+      return app
+    }
+
+    if (!clientSecret) {
+      throw new InvalidClientError('client_secret required for confidential clients')
+    }
 
     const storedSecret = this.encryptionService.decrypt(
       app.clientSecretEncrypted,
--- a/backend/nodejs/apps/tests/modules/oauth_provider/controller/oauth.provider.controller.test.ts
+++ b/backend/nodejs/apps/tests/modules/oauth_provider/controller/oauth.provider.controller.test.ts
@@ -254,6 +254,19 @@ describe('OAuthProviderController', () => {
       expect(mockRes.status.calledWith(200)).to.be.true
     })
 
+    it('should revoke successfully for a public client without a client_secret', async () => {
+      mockOAuthAppService.verifyClientCredentials.resolves({ isConfidential: false })
+      mockOAuthTokenService.revokeToken.resolves(true)
+      const req = {
+        body: { token: 'tok', client_id: 'public_cid' },
+      } as any
+
+      await controller.revoke(req, mockRes, mockNext)
+      expect(mockOAuthAppService.verifyClientCredentials.calledWith('public_cid', undefined)).to.be.true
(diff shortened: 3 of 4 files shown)


Saneajayy:
@coderabbitai review

Saneajayy:
@coderabbitai full review

Saneajayy:
@coderabbitai full review

Saneajayy:
> ✅ Action performed

@coderabbitai what's the report of the review ?

Saneajayy:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> 🧩 Analysis chain
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> set -euo pipefail
> 
> repo="pipeshub-ai/pipeshub-ai"
> pr="3232"
> 
> echo "=== Pull request state ==="
> gh pr view "$pr" --repo "$repo" --json number,title,state,isDraft,reviewDecision,comments,reviews \
>   --jq '{number,title,state,isDraft,reviewDecision, reviews: [.reviews[] | {author: .author.login, state, submittedAt, body}], comments: [.comments[] | {author: .author.login, createdAt, body}] }'
> 
> echo "=== Inline review comments ==="
> gh api "repos/$repo/pulls/$pr/comments?per_page=100" \
>   --jq '.[] | {user: .user.login, path, line, original_line, created_at, body}'
> ```
> 
> Length of output: 38210
> 
> `@Saneajayy` The latest full review completed with **no actionable comments** for the OAuth changes.
> 
> **Reviewed commit:** `2ea672c1a40128d99e8c107bccf9f42f8b802fc0` **Reviewed files:** 4 OAuth source and test files.
> 
> The review confirmed these behaviors:
> 
> * Public clients can use credential verification without `client_secret` where allowed.
> * Confidential clients still require `client_secret`.
> * Public clients are rejected for the `client_credentials` grant.
> * The changes satisfy the scope of issue `#3225`.
> 
> There are historical review findings on this PR:
> 
> 1. `Dockerfile`: Do not suppress required ARM dependency installation failures.
> 2. `scripts/start_dev.sh`: Make connector readiness checks fail on HTTP errors and use a bounded timeout.
> 3. OAuth `client_credentials`: Reject public clients. This finding is covered by the current change.
> 
> The current CodeRabbit merge-risk status is **moderate** because the connector readiness findings remain on `scripts/start_dev.sh`. GitHub also shows the PR as **review required**.
> 
> _You are interacting with an AI system._

@coderabbitai does this fix the issue #3225 ?

PR #3266, opened 2026-09-13
fix: treat AuthType.NONE toolsets as authenticated without credentials

### LINKED ISSUE #3065: Toolsets with AuthType.NONE are never treated as authenticated

Comment here to claim it. One assignee. Python toolsets path; no Kafka.

### Why this exists

A toolset with `authType: NONE` has nothing to save, so there is never a per-user credential record. Every "is this ready?" check treats a missing record as unauthenticated. The toolset stays unusable even though it needs no credentials.

### Files

- `backend/python/app/api/routes/toolsets.py` — especially the loop around line 1855 (`if not user_auth or not user_auth.get("isAuthenticated", False)`). Search the same file (and any `GET /my-toolsets` / agent tool-load path) for the same pattern.
- Tests under `backend/python/tests/unit/api/routes/` (extend `test_toolsets.py` or add a focused file).

### What to change

Treat `authType == "NONE"` (and `AuthType.NONE`) as authenticated/configured without a credential record. An agent must be able to call those tools. Add a unit test with a NONE-auth instance and no user_auth document.

### How to verify

```bash
cd backend/python
pytest tests/unit/api/routes/test_toolsets.py -q --tb=short -k none
```

(or the test file you add)

### Out of scope

Do not change OAuth or API_TOKEN flows. Do not add a fake credential write for NONE.

HarshRajSinghania:
## Summary

Toolset instances with `authType: NONE` have nothing to save, so there is never a per-user credential record. Readiness checks treated a missing record as unauthenticated, which left those toolsets unusable for agents.

Fixes #3065

## Implementation

- Added `_auth_type_is_none` / `_is_toolset_authenticated` helpers in `backend/python/app/api/routes/toolsets.py`.
- `get_authenticated_toolsets` now includes NONE-auth instances even when no user auth document exists (stores `{}` in `auth_by_instance` rather than writing a fake credential).
- `_build_toolsets_list_response` and the instance auth-status endpoint use the same helper so list/filter UIs agree.
- OAuth / API_TOKEN / BASIC_AUTH paths are unchanged.

## Testing

Added `backend/python/tests/unit/api/routes/test_toolsets_none_auth.py`:
- NONE instance + no user_auth is returned as authenticated; OAUTH without a record is not.
- Enum-shaped `authType` values with `.value == "NONE"` are treated the same as the string `"NONE"`.

Helper logic was also checked independently:

```
python3 -c "..._is_toolset_authenticated checks..."
# helper checks passed
```

I did not run the full `backend/python` pytest suite here because this environment does not have the project test dependencies installed.

Suggested verification from a checkout:

```bash
cd backend/python
pytest tests/unit/api/routes/test_toolsets_none_auth.py tests/unit/api/routes/test_toolsets.py -q --tb=short -k none
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Toolsets that do not require authentication are now correctly reported as authenticated without user credentials.
  - Toolset status and listings now consistently reflect authentication readiness.
  - Authentication types represented as enum values are handled correctly.

- **Tests**
  - Added coverage for unauthenticated toolsets and credential-dependent authentication scenarios.


shekharkadyan:
Closing as superseded by #3291 (same author, same #3065).

#3291 also covers the agent tool-load path in `agent.py`, which this PR does not. Review continues there.

PR #3280, opened 2026-09-13
fix(storage): deleting a document never removed its bytes — add a delete to every provider, and make S3 version-aware

rbales79:
## Description

`deleteDocumentById` sets `isDeleted = true` and saves the row. **Nothing in the tree ever reads
that flag back.** `grep -rn isDeleted modules/storage/` returns the single write, two
`isDeleted: false` initialisers, the schema and the type — no query, no cron, no GC, and no
`deleteObject`/`unlink` call anywhere in the module.

So a deleted document's bytes stayed in storage forever while the API returned `200`.

**It could not have done otherwise: no provider had a delete method.**
`StorageServiceInterface` declares eight operations and none of them is a delete;
`s3.provider.ts`, `local-storage.provider.ts` and `azure.provider.ts` each implement exactly
those eight. `libs/errors/storage.errors.ts` does ship a `StorageDeleteError` — an error class
for a delete that was never written.

The design intent looks like immediate delete, half-built: the endpoint exists, the error class
exists, and one connector already calls the endpoint on delete.

## What this changes

| File | Change |
|---|---|
| `services/storage.service.ts` | `deleteDocumentFromStorageService` on the interface |
| `adapter/base-storage.adapter.ts` | delegate it |
| `providers/s3.provider.ts` | implement, **version-aware** |
| `providers/local-storage.provider.ts` | implement, `unlink` via the existing `assertInsideMount` |
| `providers/azure.provider.ts` | implement, `deleteIfExists({ deleteSnapshots: 'include' })` |
| `controllers/storage.controller.ts` | call it **before** flagging the row |

Each covers the current object **and every entry in `versionHistory`**.

## The version-aware part, which is the easy bit to get wrong

On a bucket with object versioning enabled, a plain `deleteObject` writes a **delete marker** and
leaves every prior version readable. The object disappears from a normal listing and the bytes
are still there — a successful-looking delete that deleted nothing.

So the S3 provider enumerates each key with `listObjectVersions` and removes every version *and*
every delete marker, paging through `NextKeyMarker`/`NextVersionIdMarker` and filtering the
prefix down to the exact key (`Prefix` is not an exact match, so `file.pdf` would otherwise catch
`file.pdf.bak`). `deleteObjects` is chunked at its 1000-key limit.

On an **unversioned** bucket the same call returns a single entry with `VersionId: "null"`, so
this path is correct there too and needs no branch.

**This interaction is worth stating plainly:** enabling object versioning is good practice — it
turns "someone overwrote a document" from an incident into a restore — and doing that good thing
is what re-opens the hole. Two individually correct changes cancel each other, and nothing in
the system says so.

## Ordering and failure semantics

Blob first, then the flag. If storage fails, the request surfaces a 5xx and the row still says
the document exists — **which is true**. The other order produces a row recording a deletion that
did not happen.

- An object that is **already absent** is the desired end state, not a failure (`NoSuchKey` /
  `ENOENT` / `deleteIfExists` returning `succeeded: false` are all logged and skipped).
- An object that **exists and cannot be removed** raises `StorageDeleteError`.

## Tests

`npx mocha` full suite: **9828 passing, 0 failing**. `tsc --noEmit`: clean.

**8 new S3 cases** — every version of a key; delete markers as well as versions; a key that only
shares a prefix is left alone; version history as well as the current object; a key appearing
twice is deduplicated; a document with no objects succeeds with nothing to do; an already-absent
object is success; a failure raises `StorageDeleteError`.

**2 new controller cases** — storage is actually called with the document; and **the row is not
flagged when storage deletion fails**.

Also updated the existing controller fake adapter and two `StorageServiceInterface` mocks. They
failed when the interface grew, which is the mocks doing their job — worth noting for review,
since those diffs are test-support rather than new coverage.

## Verification against a running deployment

The same logic, applied to the compiled output of a running 1.x deployment, was verified
end-to-end against a **versioning-enabled** bucket:

```
before:  DELETE /api/v1/knowledgeBase/record/<id>  ->  {"success": true}
         object still present, isDeleted false on all 1001 rows

after:   0 versions, 0 delete markers, 0 readable bytes
```

**S3 (RustFS) and local disk are verified. Azure is written but not tested** — no Azure
deployment available here. The Azure path is the simplest of the three
(`deleteIfExists` with `deleteSnapshots: 'include'`), but it deserves a reviewer's eye rather
than my assurance.

## Scope

This makes the **existing** storage delete endpoint actually delete. It does **not** change the
connector record-delete path, which currently never calls that endpoint at all — that is a
separate cascade and belongs in its own PR, on top of this one.

## Summary by CodeRabbit

* **New Features**
  * Deleting a document now also removes its associated files and stored version history.
  * Supported storage providers handle missing files gracefully while reporting deletion failures.

* **Bug Fixes**
  * Prevented deleted document records from being saved when associated file removal fails.

* **Tests**
  * Added coverage for storage deletion, version history, missing objects, and error handling.


abhishek-pipeshub:
@rbales79 Can you please fix the review comments

rbales79:
Thanks for the ping, and apologies for the delay.

Before I work the review comments: this PR has collided with `8d578bb6`, which landed after I opened it. That commit added `deleteObject` to the Azure provider, the S3 provider, the base adapter and the service interface. This PR adds `deleteDocumentFromStorageService` to the same classes — different names, so GitHub reports the branch mergeable and it does merge cleanly, but the merged tree carries two parallel delete APIs on one adapter. I'd rather not hand you that.

What I read as still missing on `main`:

- `deleteDocumentById` soft-deletes only (`document.isDeleted = true`); the bytes stay. The `deleteObject` call at `storage.controller.ts:427` is the abort-a-direct-upload path, not this one.
- S3's `deleteObject` removes the current version only. On a versioned bucket that writes a delete marker and leaves every prior version in place.
- `local-storage.provider.ts` has no delete at all.

So what I'd propose is: rebase onto `main`, drop my Azure and S3 duplicates, extend your `deleteObject` to be version-aware on S3, add the local-storage implementation, and wire `deleteDocumentById` to call it.

That also resolves one of the review findings for free. CodeRabbit asked me to decode the blob name before deletion because `URL.pathname` keeps percent-encoding — true against this PR's base, but `39322ed1` added `.map(decodeURIComponent)` inside `getBlobPath`, and my Azure path already calls `getBlobPath`. Fixing it as written after a rebase would decode twice.

Happy to do that, or to keep the two APIs separate if you'd rather — your call on the shape, since it's your codebase and I'd rather not guess. The `any` escapes and the `next` assertion I'll fix either way.

PR #3286, opened 2026-09-14
fix(indexing): size chunks by tokens, not characters, against the embedder limit

### LINKED ISSUE #3285: Chunk sizing counts characters against a token limit: 50,000 chars is ~1.5x the embedder's 8,191-token ceiling
### Summary

Every block-sizing decision in the indexing path is made in **characters**, but the limit it has to respect — the embedding model's context — is expressed in **tokens**. The two are not proportional, so the trigger that decides *"this block is too large to embed as one document"* sits **past** the point at which embedding actually fails.

```python
# backend/python/app/modules/transformers/vectorstore.py
_MAX_BLOCK_CHARS_FOR_SENTENCE_SPLIT = 50_000   # :108
_OVERSIZED_CHUNK_SIZE               =  1_500   # :109

# backend/python/app/modules/parsers/text_splitting.py
MAX_TEXT_BLOCK_CHARS                = 50_000   # :23

# text-embedding-3-small / -large accept 8,191 TOKENS
```

50,000 characters is roughly **12,500 tokens** of English prose — about **1.5× the ceiling**. A 45,000-character block therefore takes the "small enough" branch at `vectorstore.py:275` and is emitted as a single `isBlock: True` document of roughly 10,000 tokens.

`grep -riE "tiktoken|token_limit|count_tokens" backend/python/app/modules/transformers/` returns nothing — there is no token counting anywhere in this path.

### Why this is not only a failed-record problem

The first-order symptom is a 400 from the provider (`Invalid 'input[N]': maximum input length is 8192 tokens`), which `is_retriable_embedding_error` classifies as non-retriable, so **the whole record fails** — losing the other 99% of a document because one paragraph was long. Observed here on 4 of 298 `.eml` records, all 3.5–4 MB.

The second-order problem is worse and outlives any truncation fix: if the input is clipped to fit, **the clipped tail is silently unretrievable**. Nothing in the index distinguishes a chunk that was embedded whole from one that lost its end. The failure stops being loud and starts being quiet.

### Three code paths can produce an over-limit document, not one

Fixing only the first leaves two open. All three are in `_build_text_documents`:

1. **`vectorstore.py:275`** — `if len(block_text) > _MAX_BLOCK_CHARS_FOR_SENTENCE_SPLIT:`. Below that threshold the block is emitted whole. Between roughly 8k and 12.5k tokens, that single document is over the limit. **This is the main defect.**

2. **`_chunk_oversized_text`** — packs sentences into ~1,500-character windows, which is safe in itself. But a sentence **longer than the window** is appended as its own chunk without being split (`if current and current_len + added_len > chunk_size` flushes `current`, then starts a new one containing the long sentence alone). A table row, a base64 payload, a minified line, or CJK text the splitter does not segment arrives as exactly this.

3. **The sentence path** — `split_into_sentences(...)` output is emitted directly as Documents, with the same single-long-sentence hole.

### Token density is the part character counts cannot approximate

Measured with `tiktoken` `cl100k_base`:

| input | chars | tokens | under the 50k char rule? | over the 8,191 token limit? |
|---|---:|---:|---|---|
| English prose | 45,000 | **10,001** | yes — not split | **yes** |
| base64-like payload | 43,200 | **27,000** | yes — not split | **yes, 3.3×** |
| prose | 49,800 | 6,641 | yes — not split | no |
| ordinary block | 47 | 10 | yes | no |

The second row is the one a character rule cannot see at any threshold: **6× the token density of the first**, inside a string that is *shorter*. Any single characters-per-token constant is wrong for one of these two.

### Suggested fix

Count tokens at the decision, and enforce the ceiling at all three emission points:

- resolve the ceiling once (`tiktoken` where available; a deliberately pessimistic `chars // 3` when not, so the fallback errs toward splitting)
- keep the character cap as a cheap pre-filter — a string shorter than `ceiling × 3` characters cannot exceed the ceiling, so the common case never pays for tokenization
- add a hard token-bounded split as the last line of defence for input the sentence splitter cannot break up
- make the limit configurable, since `text-embedding-ada-002` accepts 2,048 rather than 8,191 and a self-hosted model may differ again

**Chunk *sizing* should not change as part of this.** The ~1,500-character window and the sentence-splitting behaviour can stay exactly as they are — this is a correctness fix, not a retrieval-tuning change, and conflating the two makes the result unmeasurable.

One detail worth stating: if truncation is also applied at the embedder, **both places must read the same ceiling from the same source.** Two components disagreeing about the limit is how this class of bug returns.

### Verification

I have a working implementation of the above and can open a PR if useful.

- **Unit tests**: 15 cases covering the ceiling resolution (including malformed env overrides falling back rather than disabling the ceiling), the detection of both prose and token-dense inputs that are under the character cap, the cheap pre-filter never skipping text that is genuinely over, every split piece landing within the ceiling, and **no content being dropped** — `"".join(pieces) == text`, since truncation is the thing being fixed. All pass. There is currently no test file covering `modules/transformers/vectorstore.py`.
- **`ruff`** with the repo's own `pyproject.toml`: 72 findings on the file before, 75 after, and the only new category is `UP006` (`List` vs `list`), which already fires **45 times** on the unmodified file — so the addition matches the surrounding house style rather than introducing a new one.
- **Retrieval**: a 21-question golden set over a ~1,100-record corpus scores **recall@10 76% → 76%, zero rank flips**, MRR 0.445 → 0.452. That is the expected result and worth being explicit about: on a corpus whose largest indexed chunk is 4,475 tokens (p99 374, p50 8, **zero** over 8,191), this changes no current output. It closes a hole that today's corpus does not happen to walk through — one dense PDF or a wide spreadsheet export does.

### Environment

- `pipeshubai/pipeshub-ai:slim`
- anchors confirmed present on `main` at `23f8535`
- embedding model: `text-embedding-3-small` (8,191 tokens)

rbales79:
Fixes #3285.

### The problem

Every block-sizing constant in the indexing path bounds **characters**, but the limit that actually rejects input is the embedding model's, in **tokens**:

```
_MAX_BLOCK_CHARS_FOR_SENTENCE_SPLIT = 50_000   vectorstore.py:108
MAX_TEXT_BLOCK_CHARS                = 50_000   text_splitting.py:23
text-embedding-3-{small,large}      =  8_191 TOKENS
```

50,000 characters is roughly **12,500 tokens** of English prose — about **1.5× the ceiling**. So the trigger that decides *"too large to embed as one document"* sits **past** the point where embedding fails. A 45,000-character block takes the small-enough branch at `vectorstore.py:275` and is emitted as a single ~10,000-token document, which the provider rejects with `maximum input length is 8192 tokens`. `is_retriable_embedding_error` classifies that as non-retriable, so **the whole record fails** — losing the rest of a document because one paragraph was long.

### Three paths, not one

Fixing only the first leaves two open — all three are in `_build_text_documents`:

1. **`vectorstore.py:275`** — a block under the character cap is emitted whole. Between ~8k and ~12.5k tokens that single document is over the limit.
2. **`_chunk_oversized_text`** packs sentences into ~1,500-char windows, which is safe in itself — except a sentence **longer than the window** is appended as its own chunk unsplit. A table row, a base64 payload, a minified line, or text in a script the splitter doesn't segment arrives as exactly that.
3. **The sentence path** — `split_into_sentences` output is emitted directly, with the same hole.

### Why a bigger constant would not have fixed it

Measured with `tiktoken` `cl100k_base`:

| input | chars | tokens | under 50k? | over 8,191? |
|---|---:|---:|---|---|
| English prose | 45,000 | **10,001** | yes | **yes** |
| base64 payload | 43,200 | **27,000** | yes | **yes, 3.3×** |
| English prose | 49,800 | 6,641 | yes | no |
| ordinary block | 47 | 10 | yes | no |

**The base64 row is 6× the token density of the first, in a shorter string.** Any single characters-per-token constant is wrong for one of them — which is why the unit has to change rather than the number.

### Implementation notes

- **The character cap is kept as a cheap pre-filter, not removed.** A string shorter than `ceiling × 3` characters cannot exceed the ceiling, so the common case never pays for tokenization.
- **`tiktoken` is imported optionally.** When unavailable the fallback estimates at 3 chars/token, which *over*-estimates and therefore errs toward splitting rather than toward emitting something that will be rejected.
- **The limit is configurable** via `PIPESHUB_EMBED_TOKEN_LIMIT`, since `ada-002` accepts 2,048 rather than 8,191 and self-hosted models vary. A malformed value falls back to the default rather than silently disabling the ceiling.
- **Chunk *sizing* is deliberately unchanged.** The ~1,500-character window and the sentence-splitting behaviour are untouched. This is a correctness fix, not a retrieval-tuning change, and conflating the two would make the result unmeasurable.

If truncation is ever also added at the embedder, both places should read the ceiling from this same source — two components disagreeing about the limit is how this class of bug returns.

### Tests

15 cases in `tests/unit/modules/transformers/test_vectorstore_token_sizing.py`. They encode the **regression**, not the implementation:

- each over-limit case asserts `len(text) < 50_000` *alongside* the token assertion, so the test states *"under the character cap AND over the token ceiling"* — which is the bug
- one asserts `"".join(pieces) == text`, because a split that silently dropped content would otherwise pass while reintroducing the problem this fixes
- malformed env overrides must **fall back**, not disable the ceiling
- the cheap pre-filter is asserted never to skip text that is genuinely over

`test_vectorstore.py` already covers `_chunk_oversized_text` and `_build_text_documents`, but nothing in the existing suite asserts any token bound — `grep -r "8191\|8192\|token_limit\|tiktoken" tests/unit/modules/transformers/` returns nothing before this change.

### Verification

**Existing tests are unaffected.** I could not get the vectorstore suite fully green in my environment — 158 tests fail there for reasons unrelated to this change — so the comparison is baseline-relative and stated as such:

| | failed | passed |
|---|---:|---:|
| unmodified `main` | 158 | 62 |
| with this change | **158** | **77** |

The failure sets are **identical** (`comm` on the sorted `FAILED` lines shows zero added, zero removed), and the pass delta is exactly the 15 tests added here.

**Retrieval:** a 21-question golden set over a ~1,100-record corpus scores **recall@10 76% → 76%, zero rank flips**, MRR 0.445 → 0.452. That is the expected result and worth being explicit about — on a corpus whose largest indexed chunk is 4,475 tokens (p99 374, p50 8, **zero** over 8,191), this changes no current output. It closes a hole that corpus does not happen to walk through; one dense PDF or a wide spreadsheet export does.

**Lint:** `ruff` with the repo's `pyproject.toml` reports 72 findings on `vectorstore.py` before and 75 after; the only new category is `UP006` (`List` vs `list`), which already fires 45 times on the unmodified file, so the addition follows the surrounding style rather than introducing a new one.

Branched from `main` at `23f8535`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved embedding reliability by enforcing model token limits before processing text.
  * Automatically splits oversized text while preserving complete content and UTF-8 characters.
  * Handles token-dense content, long unbroken strings, and individual sentences more reliably.
  * Supports configurable token limits with safe fallback behavior when settings or optional tokenization support are unavailable.
  * Prevents oversized text chunks from being sent to the embedding service.


rbales79:
Thanks — three of these were correct and are fixed in `c9c30e2`. Measurements rather than assertions, and one of them was a genuine bug in the patch.

### 1. The fallback was not a bound — correct, and worse than described

The estimate was 3 characters per token and the docstring called it *"deliberately pessimistic"*. That is true only for ASCII prose. Measured against `cl100k_base`:

| input | chars | bytes | tokens | chars/token | `len/3` estimate |
|---|---:|---:|---:|---:|---|
| ASCII prose | 1,200 | 1,200 | 201 | 5.97 | 400 — safe |
| CJK | 1,100 | 3,300 | 1,000 | 1.10 | 366 — **under by 2.7×** |
| emoji | 100 | 400 | 300 | 0.33 | 33 — **under by 9×** |
| base64 | 1,600 | 1,600 | 900 | 1.78 | 533 — **under by 1.7×** |

It under-counted on **three of four** input classes. A BPE token encodes at least one byte, so the **UTF-8 byte length is a guaranteed upper bound** for any input. `_token_len` now uses it.

**The same error was in the cheap pre-filter, and worse there** — it compared *characters* against the ceiling. A character count is not an upper bound on tokens, so the guard waved through exactly the dense input the function exists to catch. It now measures bytes, with an `isascii()` fast path where the two coincide.

### 2. UTF-8 boundaries — correct, and the test had a hole

Confirmed directly: 4,600 tokens of CJK and emoji split into 92 pieces produced **20 replacement characters**, and the pieces did **not** rejoin to the source.

Slicing now goes through `decode_bytes` plus an incremental decoder, which carries an incomplete trailing sequence into the next piece.

Worth stating plainly: my `"".join(pieces) == text` test passed only because the fixture was ASCII, where this failure cannot occur. That was a hole in the test, not luck.

**And the pieces are now verified rather than assumed.** Two effects make a slice of `ceiling` units come back *over* `ceiling` tokens — carrying a partial character forward adds bytes to the next piece, and re-encoding a decoded slice does not always reproduce its token count because BPE merges differ once the text is cut. Both are small; small is not a bound. Each attempt is measured and the step halved until every piece fits.

### 3 & 4. Test fixtures — both correct

- The ceiling is now a fixed local constant, so an exported `PIPESHUB_EMBED_TOKEN_LIMIT` cannot move the fixtures out from under the assertions.
- **`tiktoken` is not a declared dependency** — I verified it is absent from `pyproject.toml` while being present in the runtime image, so both paths are real. Tests needing an encoder now skip without one, and a fixture forces the fallback path so it is exercised even where `tiktoken` *is* installed.

27 tests, up from 15, with losslessness asserted across CJK, emoji and Latin-extended input and an explicit assertion that no `U+FFFD` appears.

### 5. Routing the limit through `ConfigurationService` — declining, with a reason

`_build_text_documents`, `_exceeds_token_ceiling` and `_split_to_token_ceiling` are module-level functions with no access to `config_service`, and `get_config()` is async. Threading it through would mean changing those signatures and their call sites — a refactor considerably larger than this fix, and one that would make the correctness change harder to review on its own.

It would also be inconsistent with the surrounding file, which reads four other settings directly at module scope:

```
vectorstore.py:92   os.getenv("EMBEDDING_BATCH_CONCURRENCY")
vectorstore.py:192  os.getenv("EMBED_SENTENCE_MIN_WORDS", ...)
vectorstore.py:458  os.getenv("VECTOR_STORAGE_RECONCILE_ENABLED", ...)
```

`PIPESHUB_EMBED_TOKEN_LIMIT` follows the same pattern as its immediate neighbours. Happy to move all four to `ConfigurationService` in a separate PR if that is the direction — it seems like a decision about the file rather than about this change.

### Verification after the fixes

- **27 tests pass** (15 → 27)
- **Existing vectorstore suite unaffected**: identical failure set before and after (`comm` on sorted `FAILED` lines: 0 added, 0 removed), pass count up by exactly the tests added. I still cannot run that suite green locally — 158 fail there on unmodified `main` too — so it stays baseline-relative.
- **`ruff`**: 72 findings before, 74 after; the delta is +3 `UP006` (`List`, matching the 45 already in the file) and an `I001` import-order finding that my changes *removed*.

abhishek-pipeshub (approved):

PR #3297, opened 2026-09-15
fix(auth): publish the Azure AD provider config under the key the frontend reads

rbales79:
## Problem

A correctly configured Azure AD login renders **"Microsoft sign-in is not configured for your organisation"**, with no error logged anywhere.

The backend publishes the provider config for the login page under the lowercase key `azuread`, in two places:

- `userAccount.controller.ts:325` — the `configMethodMap` entry (`key: 'azuread'`), preserved by a no-op ternary at `:364`
- `userAccount.controller.ts:1560` — the multi-step branch (`authProviders.azuread = cfg.data`)

The login form reads the camelCase key: `frontend/app/(public)/forms/single-provider.tsx:280` — `authProviders?.azureAd?.clientId`. The same controller already uses `'azureAd'` as the method key on the JIT path (`:1487`), so the backend disagrees with itself.

## Reproduction

Fresh deployment, Azure AD configured via `configurationManager/authConfig/azureAd`, method enabled via `orgAuthConfig/updateAuthMethod`:

```
POST /api/v1/userAccount/initAuth
→ {"allowedMethods":["password","azureAd"],"authProviders":{"azuread":{"clientId":…}}}
```

The Microsoft button appears (from `allowedMethods`) and is dead (the form finds no `authProviders.azureAd`).

## Fix

Three string edits, no behaviour change for any other provider. After the change the same request returns `authProviders.azureAd` and the flow completes.

## Testing

Reproduced and verified on a running deployment as above. No test added: the controller's provider map is not unit-tested today and the change is a key rename; happy to add one if you point me at the pattern you'd like.

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized Azure AD configuration and authentication responses to consistently use the `azureAd` key during initial and multi-step authentication flows.


abhishekgml1991 (dismissed):

abhishek-pipeshub:
@rbales79 Can you please fix the test cases

rbales79:
Thanks — fixed in the commit just pushed.

The failing assertion was encoding the bug this PR fixes. `userAccount.controller.test.ts:2539` expected `authProviders` to have property `azuread` (all lowercase), but the frontend reads `authProviders?.azureAd?.clientId` / `?.authority` (`frontend/app/(public)/forms/single-provider.tsx:280,282`). Since the API must publish the key the frontend reads, the assertion now expects `azureAd`. That was the only remaining lowercase `azuread` left in `src/` or `tests/` after the source change.

One line, test-only:

```diff
-        expect(jsonArg.authProviders).to.have.property('azuread');
+        expect(jsonArg.authProviders).to.have.property('azureAd');
```

Verified locally with `TS_NODE_PROJECT=tsconfig.test.json mocha`: the targeted test passes, and the full suite is **9703 passing, 4 pending**, exit 0.

The `redis-cluster-integration-tests` failure was the same single test — that job runs the whole Node suite, and nothing Redis-related failed in it (the cluster came up healthy, "All nodes agree about slots configuration"). It should go green with this commit too.

abhishek-pipeshub (approved):

PR #3312, opened 2026-09-16
fix(integration-tests): NameError on Tuple in local_auth.py

siddhant-ota:
## Summary

#3289 changed `_authenticate` in `integration-tests/helper/local_auth.py` to return `Tuple[str, str]`, but the file does not import `Tuple`. Importing the module raises `NameError: name 'Tuple' is not defined`.

`integration-tests/conftest.py` imports `local_auth`, so in the Integration Tests workflow:

1. "Run helper unit tests" fails at conftest load.
2. The service start, wait, and integration test steps are skipped.
3. The Playwright step still runs and every test fails with `net::ERR_CONNECTION_REFUSED` at `http://localhost:3000/login/`.

Every PR based on main after #3289 fails this way, for example run [35123340808](https://github.com/pipeshub-ai/pipeshub-ai/actions/runs/35123340808) on #3311.

The fix uses the built-in `tuple[str, str]`, the same as the other annotations in the file.

## Test plan

- [x] `pytest unit/` in `integration-tests/`: `NameError` before the change, 4 passed after.
- [x] `ruff check helper/local_auth.py` passes.

## Summary by CodeRabbit

- **Refactor**
  - Modernized internal type annotation syntax without changing functionality.


siddhant-ota:
Folded into #3311.

PR #3316, opened 2026-09-17
ci: keep the PR-validation checkout token out of reach of pull-request code

shekharkadyan:
## In one line

The pull-request validation jobs leave their access token in a place the pull request's own code could read; this stops that.

## Background

`.github/workflows/main.yml` (PR Validations) runs on every pull request. Each of its jobs starts with `actions/checkout@v4`, which by default writes the workflow's `GITHUB_TOKEN` into the checked-out repository's local Git config (`.git/config`).

Each job then runs code that the pull request author can change — a build, the test files themselves, or, in the connector-coverage job, `scripts/connector_coverage.py`. That code runs in the same workspace, so it can read the token out of `.git/config`. On a public repository, anyone can open a pull request, so this is reachable by an outside contributor.

The token here is read-only (`contents: read` at the top of the file), which limits what could be done with it — but there is no reason to leave it readable at all.

## What changed

Every checkout step in this workflow now sets `persist-credentials: false`, so the token is used for the checkout and not written to disk afterwards.

```yaml
      - name: Checkout code
        uses: actions/checkout@v4
        with:
          persist-credentials: false
```

No job in this workflow pushes, commits or tags — verified by searching the file for any `git push`, `git commit`, `git tag`, `gh release`, or commit action — so none of them needs the persisted credential and nothing breaks.

## Why all seven, not one

A review of the merged #3176 flagged this on the connector-coverage job it added. The same checkout pattern is on the other six jobs in the same pull-request workflow, and each equally runs pull-request-controlled code with the same read-only token. Fixing one and leaving six identical holes would be a false sense of security, and the change is safe everywhere here because no job pushes.

## How to test it

There is nothing to run. Confirm the setting is present on every checkout:

```bash
grep -A1 'actions/checkout@v4' .github/workflows/main.yml
```

Each should be followed by a `with:` block containing `persist-credentials: false`. The jobs behave exactly as before — they still check out the code and run the same steps.

## Risk

Workflow-only, and hardening-only. Nothing a job does changes; the token is simply not left on disk. Reverting this would restore the previous behaviour with no other effect.

https://claude.ai/code/session_01HpNmn9X69EiiBEDGYTmLWZ

## Summary by CodeRabbit

- **Chores**
  - Improved CI workflow security by preventing temporary repository access credentials from being saved after checkout.
  - Applied this protection consistently across validation, linting, type-checking, Docker Compose, workflow, and coverage checks.
  - No changes to application functionality or user-facing behavior.


shekharkadyan:
Two updates.

The four jobs that run `git fetch origin main` after checkout — `validate-backend`, `ruff-lint-pr`, `pyright-type-check-pr`, `validate-docker-compose` — go green on this PR's own run, which executes them with `persist-credentials: false` in place. That confirms the fetch still works: on a public repository the unauthenticated HTTPS fetch of `main` succeeds without the token that checkout used to leave in `.git/config`. No job needed the persisted credential.

The explanation that was repeated on all seven checkout steps now appears once, above the `permissions:` block, and each checkout keeps only the `persist-credentials: false` line.

shekhar-pipeshub (approved):

PR #3351, opened 2026-09-19
fix: delete records by key in the Dropbox, BookStack and Salesforce handlers

shekharkadyan:
## What was wrong

`DataSourceEntitiesProcessor.on_record_deleted(record_id)` looks the record up by its **key**. Four connectors passed the source system's id instead (`record_update.external_record_id`):

- Dropbox (team): `dropbox/connector.py`
- Dropbox (individual): `dropbox_individual/connector.py`
- BookStack: `bookstack/connector.py`
- Salesforce: `salesforce/connector.py`

With the source id, the lookup finds nothing, so the delete does nothing and the record stays indexed. SharePoint had the same bug; it was fixed in #3347.

## The fix

Each handler now does what SharePoint, OneDrive, Box and Nextcloud do. It looks the record up with `get_record_by_external_id(connector_id, external_record_id)` and deletes it by `record.id`. An item that was never indexed is skipped. The external id in these updates has the same form ingestion stores (`page/<id>` for BookStack, the Dropbox entry id for Dropbox), so the lookup matches the records that ingestion created.

## Reachable today?

**None of these four paths runs today.** Nothing in these connectors currently produces an update with `is_deleted=True`:

- **Dropbox, team and individual:** the code that would build a deleted update from `DeletedMetadata` is commented out.
- **BookStack and Salesforce:** nothing sets `is_deleted=True` anywhere.

This PR doesn't turn deletion detection on; that is a separate decision. It makes the handlers correct, so enabling detection later doesn't ship a delete that silently does nothing.

I checked every other `on_record_deleted(` call under `app/connectors/sources`. They all pass the internal key already: Gmail, Google Drive, OneDrive, SharePoint, Box, Nextcloud, Notion, GitHub, GitLab, Postgres, MariaDB, Snowflake, Azure Files and Web.

## Tests

- Nine existing tests asserted the wrong call (`on_record_deleted(record_id=<source id>)`), or only that it was called. They now assert the lookup by source id and the delete by record key.
- New `test_deleted_record_never_indexed` tests for BookStack, Dropbox, Dropbox individual and Salesforce check that an unknown item deletes nothing.
- All 1,878 tests in the BookStack, Dropbox and Salesforce unit suites pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deletion handling for BookStack, Dropbox, Dropbox Individual, and Salesforce records.
  * Deleted records are matched to existing records before removal, preventing incorrect deletions.
  * Deletion actions are skipped when no matching record exists.

* **Tests**
  * Expanded coverage for record deletion, filtering, permissions, synchronization, and connector workflows.


shekharkadyan:
On CodeRabbit's two outside-diff findings (`test_deep_sync_bookstack.py:315-325` and `test_bookstack_connector.py:687-695`), both hold. Fixed in 869d3514e.

After this PR, a delete runs only when the lookup finds a record, and those tests' fixtures return `None`, so the raising `on_record_deleted` was never reached. The same thing had happened to three more tests of the same kind:
- `test_bookstack_connector_comprehensive.py::test_exception_handled`
- `test_dropbox_connector.py::test_exception_handled`
- `test_dropbox_connector_extended.py::test_error_handling`

A sixth, `test_dropbox_individual_extended.py::test_exception_swallowed`, is fixed as well. All six now make the lookup return `MagicMock(id="rec-key")` and assert `on_record_deleted` was awaited with `record_id="rec-key"`, so each one proves the error path ran and was handled. All 1,878 BookStack, Dropbox and Salesforce unit tests pass.

shekharkadyan:
On CodeRabbit's outside-diff finding (`test_dropbox_connector.py:1248-1262`), it holds. The failing delete there is set up over several lines, which my earlier sweep missed. Fixed in bfaebe5e7, together with the one remaining test of the same kind, `test_salesforce_connector.py::TestHandleRecordUpdatesRemaining::test_logs_and_swallows_processing_errors`. Both now resolve `MagicMock(id="rec-key")` and assert `on_record_deleted` was awaited with it. A repo-wide search for `on_record_deleted = AsyncMock(` with a side effect in these suites finds no others. All 1,878 BookStack, Dropbox and Salesforce unit tests pass.

shekhar-pipeshub (approved):

PR #3356, opened 2026-09-19
ci: leave placeholder connectors out of the integration coverage count

shekharkadyan:
## What was wrong

We keep a running count of which connectors have integration tests, meaning tests that sync against the real service. CI uses it to block a new connector from shipping without one. The count included seven entries that aren't working connectors: Calendar, Meet, Forms, Slides, Docs, Zendesk and Airtable. They're placeholders, registered so they can appear in the product, but syncing is switched off and there's nothing behind them to sync.

## What it meant

The coverage figure looked worse than reality: 27 of 50 connectors, when it's really 27 of 43. The seven placeholders sat permanently on the "untested" list, burying the real gaps. Anyone asking which connector needs a test next had to know to ignore them.

## The fix, and why this way

The count now leaves out any connector that is declared with syncing switched off. I used that rule rather than a hard-coded list of seven names. If one of these placeholders is ever built out and syncing is switched on, it rejoins the count automatically and CI asks for a test. Nobody has to remember to update a list.

One detail needed care. The file that defines the placeholders also defines placeholder versions of Linear and Slack, but the product uses the real Linear and Slack connectors from elsewhere. So the check follows the class the product actually loads, not the name, and Linear and Slack stay counted.

## What changes going forward

- The report shows the true figure, 27 of 43 (63%), and the untested list holds only real connectors.
- A placeholder that gets a real sync shows up as untested straight away.
- The script has its own tests, run in CI before the coverage check. They confirm the seven placeholders are left out, and that Linear and Slack are still counted.

## How it was checked

The report now shows 27/43 with none of the seven placeholders listed, and the CI coverage check passes. The new tests pass. One of them uses a small made-up setup where a placeholder and a real connector share a name, and confirms only the placeholder is dropped.

## Summary by CodeRabbit

- **Tests**
  - Connector coverage reporting now excludes placeholder connectors that do not support synchronization.
  - Updated the integration-coverage baseline to reflect the revised connector scope.
  - Seven connectors are no longer listed as uncovered, while the number of covered connectors remains unchanged.
  - Added automated tests to verify connector registration and coverage behavior.
  - Coverage validation now runs automatically as part of the continuous integration workflow.


shekharkadyan:
On CodeRabbit's nitpick about regression coverage for `registered_connectors()`: added in 475ed1695. `scripts/test_connector_coverage.py` (stdlib `unittest`, so the job needs nothing installed) checks two things. Against this repo, the seven sync-less placeholders are skipped while `linear`, `slack` and `slackworkspace` are counted. With a small synthetic factory where a placeholder and a real class share a name, only the placeholder the factory imports is dropped. The `connector-coverage` job in `main.yml` runs it before `--check`.

shekhar-pipeshub (approved):

PR #3365, opened 2026-09-19
test: check what an ordinary member can see, with realistic files

shekharkadyan:
## What was wrong

Almost every search test signs in as the org admin, and the admin can see everything. So if a sharing rule were ignored (a collection shared with a team, a person removed from that team, a whole-org share taken back), every test would still pass. Only one small test checked search as an ordinary member, and it covered a single case: sharing directly with one person.

The files we indexed were also nothing like real ones. They ranged from 12 bytes to about 1 KB, had plain English names, and sat at the top level. Real collections hold multi-megabyte exports, spreadsheets with several sheets, long Word documents, PDFs, CSVs with awkward quoting, non-English names and folders nested several levels deep. Each of those takes a different route through parsing and indexing, and none of those routes was tested.

## What it meant

A permissions leak could ship without any test noticing. Examples: someone still finding a collection after being removed from the team it was shared with, or a deleted team's access staying behind. We also could not say whether the second sheet of a spreadsheet, the last page of a PDF, or the middle of a large export could actually be found.

Writing this uncovered a real bug. When a file is uploaded without the extra path information, a non-English name gets garbled: "résumé" becomes "rÃ©sumÃ©". The main knowledge-base page happens to send the path, so it is not affected. The Projects page upload and any API client do not send it, so they are. The cause is that the upload middleware (multer) decodes file names as Latin-1 by default. I confirmed this with multer on its own, using the same version and settings.

## The fix, and why this way

**A table of sharing shapes, checked as the person.** A new test creates a real non-admin member and walks through a list of cases:

- the collection is not shared at all;
- it is shared with them, then unshared;
- it is shared with their team, then they leave the team;
- it is shared with their team, then the team's share is removed;
- it is shared with their team, then the team is deleted;
- it is shared with a team they are not in;
- it is shared with a team they join later;
- it is shared with the whole org, then unshared.

For each case, the test checks what that person can actually do: search, open the collection, and open a file. Each check runs on the very next request after the change, with no waiting. Adding a case means adding one row.

A deleted file is also checked: it disappears from the member's search straight away, and the rest of the collection stays visible.

**Realistic files, generated at test time.** A new helper builds the files from a fixed seed, so every run gets the same bytes and nothing large is committed:

- a 3 MB export (the size can be set with `PIPESHUB_IT_LARGE_FILE_MB`);
- a three-sheet workbook;
- a six-page Word document with a table;
- a five-page PDF;
- a CSV with quoted commas and line breaks inside cells;
- a file with a name in several scripts and emoji;
- a file five folders deep with non-English folder names.

Each file contains a made-up word that appears nowhere else. The workbook has one per sheet, and the Word document and PDF have theirs on the last page. That lets a test say *which* file a search found, not just that something came back. It also catches a parser that reads only the first sheet or the first page.

The files are uploaded the way the knowledge-base page uploads them. The nested file therefore also exercises the path where a folder upload creates its own folders.

The "can see" checks require every one of those words to find its own file, searched as the member. The "cannot see" checks require zero results, which depends only on the permission filter and not on how search ranks results.

**Connectors too.** The Web connector is shared with the whole team, so a new check confirms an ordinary member finds a crawled page. Local FS is a personal connector, so another check confirms the owner finds their file and another member does not.

**The name bug is recorded, not fixed here.** Fixing it is a one-line change: set `defParamCharset: 'utf8'` where multer is configured in `libs/middlewares/file_processor`. That changes how the product stores names, so it belongs in its own PR. For now there is a test for the correct behaviour, marked as an expected failure. It is marked strict, so the day the bug is fixed the test turns red, and someone has to remove the marker.

## What changes going forward

- A permissions change that lets the wrong person see a collection, or keeps a removed person's access, now fails CI. This runs on both Neo4j and ArangoDB.
- A file type or size that stops indexing, loses its name, or has part of its content become unfindable fails its own named test, such as `test_every_part_of_it_can_be_found[xlsx]`. It does not also turn every permission test red.
- A new `permissions` marker lets the whole set be run on its own: `-m permissions`, or `permissions` in the workflow's manual-run menu.

Gaps this does not close: there is no second organisation in the test stack, so nothing checks that data stays inside one org. The connectors that can carry real per-user permissions (Drive, SharePoint, Confluence, Jira) need outside accounts and are not covered here.

## How it was checked

- The 12 new unit tests pass locally. They read each generated file back with an ordinary library (openpyxl, python-docx, pdfplumber) to confirm every word is where it should be and appears in only one file.
- The integration tests collect cleanly under `-m permissions`: 31 in the matrix module and one each in the Web and Local FS suites.
- The garbled-name bug was reproduced locally against multer 2.2.0 and 2.3.0.
- The integration tests themselves need the full stack and an AI key, so they run in CI. A focused run (only `permissions`, on both graph stores) is waiting for environment approval: https://github.com/pipeshub-ai/pipeshub-ai/actions/runs/35434585805

## Summary by CodeRabbit

* **Tests**
  * Added coverage for knowledge-base sharing permissions, including individual, team, and organization access.
  * Added validation that revoked, removed, or deleted access prevents content discovery.
  * Added connector visibility checks for local files and crawled web content.
  * Added coverage for realistic uploads, including large, multilingual, nested, spreadsheet, document, PDF, and CSV files.
  * Added checks for indexing completion, searchability, file metadata, and access restrictions.


shekharkadyan:
Merged the latest main into this branch.

- **Workflow conflict:** main added `resilience` to the manual-run menu in the same place this PR adds `permissions`. Both are kept. The rest of main's changes merged without conflicts.
- **Expected-failure marker removed:** #3370 now makes the upload middleware read file names as UTF-8. The knowledge-base upload route builds its upload handler through that change, so `test_a_non_ascii_name_survives_an_upload_without_a_path` is now a normal test. It should pass, and it will fail if non-English names get garbled again.

Checks after the merge:

- The 12 unit tests pass.
- `-m permissions` still selects the same 33 tests.
- actionlint reports the same five shellcheck notices it reports on main, and nothing new.

shekhar-pipeshub (approved):

PR #3367, opened 2026-09-19
Test S3, Azure and MongoDB replica sets against real servers, and fix two Azure bugs

shekharkadyan:
## What was wrong

PipesHub can store files on local disk, Amazon S3 or Azure Blob Storage, and can run MongoDB as a single server or as a replica set, which is what Kubernetes (Helm) installs use. Every automated test that touches a real server uses local disk and a single MongoDB server. The S3 and Azure code had only been tested against pretend versions of the AWS and Azure libraries. The code that runs only on a replica set had never met a real database: that is the many writes wrapped in transactions, such as the rule that an organization always keeps at least one admin.

Running the Azure code against a real storage server for the first time turned up two bugs:

- **Every Azure download link failed.** Creating a signed link threw "Invalid permission: [". The Azure library reads the requested permissions from their text form, and the code passed a plain object, whose text form is "[object Object]". Downloads and large-file uploads both depend on these links.
- **Files with a space or a non-English character in their name got lost.** After upload, the adapter looked the file up under its web-encoded name ("Quarterly%20report.pdf" instead of "Quarterly report.pdf"), which doesn't exist. Reading it failed, and editing it quietly wrote a second, separate file.

## What it meant

A customer storing files on Azure could not download any file, and any file named like "Q3 report.pdf" was unreadable after upload. None of our tests could notice, because they mocked the Azure library. The same blind spot applied to S3 and to replica-set MongoDB: a bug on those setups would first be found by a customer running them.

## The fix, and why this way

The two Azure bugs are fixed in the adapter. It now builds permissions with the library's own helper. It also decodes the name in a stored link before looking the file up, for both the normal Azure link form and the form emulators use. Unit tests pin both fixes.

A new workflow, "Storage backends", runs the real adapters against real servers. It uses two free stand-ins that behave like the cloud services: MinIO speaks the S3 protocol, and Azurite is Microsoft's own emulator for Azure Storage. It checks:

- storing and reading back small, 12 MB and oddly named files;
- updates and older versions;
- signed download links, including the download file name;
- browser uploads to a signed link, and S3 multi-part uploads;
- that a missing file is reported as an error, not as empty content.

The same workflow runs the MongoDB code against a replica set and against a single server. It checks that transactions really roll back, and that the admin rules hold. The one-admin minimum and the five-admin maximum must survive two changes arriving at the same moment. I confirmed the concurrency test would catch the bug it's meant to catch. Turning off the safeguard that serializes those checks made it fail on every run.

These are focused tests of the storage layer, not a full product install. That keeps them to a few minutes on a standard GitHub machine, with no secrets. It also means a failure points at the storage code directly. The S3 adapter has no setting for a custom server address, so the test points the AWS library at MinIO from inside the test process. The product code is unchanged for S3.

## What changes going forward

Changes to the storage code, the MongoDB connection or the admin rules now run against real S3-compatible storage, Azure storage and both MongoDB setups before they merge. The whole set also runs nightly, so an update to the AWS or Azure libraries that breaks something shows up within a day.

## How it was checked

- The workflow's own start-up steps and commands were run locally with the same images and ports: S3 13/13, Azure 13/13, MongoDB replica set 8/8, and single server 7/7. The replica-set run was repeated five times to check it isn't flaky.
- Before the fix, the Azure run failed 10 of 13 checks.
- The Azure adapter's unit tests pass (86), including four new ones.
- actionlint passes on the workflow.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Azure Blob Storage signed URLs and direct uploads with correct read and write permissions.
  * Improved handling of encoded blob names and supported Azure URL formats.
  * Added validation to prevent access to blobs from unrelated containers.

* **Tests**
  * Added integration coverage for S3, Azure Blob Storage, and MongoDB configurations, including uploads, transactions, and concurrency scenarios.

* **Documentation**
  * Added setup and execution guidance for storage and MongoDB integration tests.


shekharkadyan:
@coderabbitai Thanks, the finding about the container check in the Azure adapter's constructor (around line 91) holds, and it's fixed in 8bcda4cb3.

The constructor started creating the container but never waited for it. The first upload could reach Azure before the container existed, and a failed check became an unhandled rejection. The Azurite run didn't show it because the create request goes out first and the emulator answers in order. Real Azure doesn't promise that.

What changed:
- The constructor keeps the check as a promise, with a handler attached so a failure isn't unhandled.
- Every blob operation (upload, update, read, signed link, direct-upload link) waits for that promise inside its existing `try`. A failure comes back as that operation's `StorageConfigurationError`.
- A failed check is retried once on the next operation, so a brief outage at start-up doesn't break the adapter for good.

Tests:
- Three new unit tests: the first upload waits for the container; a failed check is reported by the operation, with no unhandled rejection; a check that fails once, then succeeds, recovers. The first two fail on the old code.
- A new Azurite test uploads straight away through a fresh adapter to a container that doesn't exist yet.
- All 623 storage unit tests pass, and the Azure suite against Azurite passes 14/14.

shekhar-pipeshub (approved):

PR #3381, opened 2026-09-19
fix: say a language model is missing instead of failing records with "'llm'"

shekharkadyan:
## What was wrong

When an organisation has no language model (LLM) set up, indexing can't finish. Every uploaded or synced file failed, and the reason stored on each record read:

> Failed to process document: 'llm'

That text is the raw name of a missing settings key. The code looked up the model settings with a lookup that crashes when the entry isn't there, and the crash message became the failure reason shown to users.

The indexing documentation added to the confusion. It described the language model as used only in the final "enrichment" step, after a document is already searchable. In practice indexing also uses it earlier: for images, spreadsheets, CSVs and tables inside documents.

## What it meant

An admin who set up embeddings but no language model saw every document fail with a message that doesn't say what is wrong or how to fix it. The docs suggested a missing language model shouldn't stop indexing at all, so the admin had nowhere to look. A performance benchmark ran into exactly this: all 20 test files failed with "'llm'".

## The fix, and why this way

- The model lookup now reads the settings safely. When no language model is configured, it raises a named error, `LLMNotConfiguredError`, whose message says so and where to add one (Workspace > AI Models).
- That error is a kind of `ValueError`, which is what the lookup already raised when the model list was present but empty, so existing error handling keeps working unchanged. The message reaches the record's failure reason through the existing path, so nothing new is needed in the UI.
- `docs/indexing-service.md` now states that indexing needs a language model and lists where it is used before a record can finish.

The fix sits in the one shared lookup that indexing, chat and the parsers all go through, so every caller gets the clearer message.

## What changes going forward

With no model configured, a failed record now reads "No language model is configured. Add one under Workspace > AI Models; indexing and chat both need it." That tells the admin what to do. Nothing changes when a model is configured.

## How it was checked

- New unit tests:
  - With no model entry at all, with no settings, or with only an embedding model, the lookup raises the new error, not a `KeyError`.
  - The lookup indexing uses (the "indexing" model role) reaches the same error.
  - A spreadsheet processed with no model configured fails with the new message, and the message does not contain `'llm'`.
- All seven new or updated tests fail against `main` and pass with the fix.
- Every unit-test file that touches the model lookup was run: 1,038 passed, 27 skipped, 4 failed. The same 4 fail on a clean `main` checkout in this environment (an embedding-provider test and three agent-route tests), so they are not caused by this change. Two VLM OCR test files couldn't load here because `langchain` isn't installed locally.
- ruff shows no new findings on the changed files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling when no language model is configured.
  - Administrators now receive clear setup guidance directing them to Workspace > AI Models.
  - Chat and document processing display user-friendly configuration errors instead of technical errors.
  - Record processing failures clearly identify missing language model configuration.

- **Documentation**
  - Documented language model requirements for image, spreadsheet, CSV, table, and inline enrichment processing.


shekharkadyan:
@cursor thanks, that held. Fixed in 39c2e1e94.

- `get_model_config` in `app/api/routes/chatbot.py`, which chat, agents and the parse-file tool all use, now reads the AI-models config safely: a missing config or a missing `llm` bucket counts as empty, including on the refresh path for a model key. It raises the same `LLMNotConfiguredError` as the indexing lookup, so the message lives in one place.
- `get_llm_for_chat` passes that error through instead of wrapping it as "Failed to initialize LLM: …". Chat and agent chat report errors as a stream event once the response has started, and that convention is unchanged. The event now says "No language model is configured. Add one under Workspace > AI Models; …" instead of `Failed to initialize LLM: 'llm'`. Every other failure is still wrapped as before.
- The parse-file tool's token check now reports "Token check failed: No language model is configured…" instead of "Token check failed: 'llm'".
- No other stored-config `["llm"]` lookups are left in `backend/python/app`. The retrieval service already used `.get`. On the Node side, `aiConfig.llm.length` in `cm_controller.ts` reads the request body of an admin saving the AI-models settings, not the stored config, so I left it out of this PR.

Tests: `get_model_config` with no config, no `llm` bucket, or embeddings only, plus the refresh path; `get_llm_for_chat` keeping the clear message; and `check_token_limit` running the real `get_model_config` with no LLM. All of them fail against the previous commit. Four existing tests that pinned the old "No LLM configurations found" or wrapped message now expect the new error. The chat, agent, parse-file and LLM suites give 811 passed and 8 failed; the same 8 agent-route tests fail on `main` in this environment. ruff shows no new findings.

shekharkadyan:
@coderabbitai thanks, fixed in b5ecff00a: the docstring at `test_chatbot_97_coverage.py:163` now says the test expects `LLMNotConfiguredError`. No other docstrings next to the updated assertions mentioned the old error. The file passes and ruff shows no new findings.

shekhar-pipeshub (approved):

PR #3430, opened 2026-09-20
fix(api-docs): let the error envelope carry the request id it already sends

shekharkadyan:
## What was wrong

Every error the Node API returns carries a `requestId`, so someone reporting a
problem can quote it and it can be found in the logs. The OpenAPI spec was never
updated to say so, and its error schema sets `additionalProperties: false` —
meaning any field it does not list makes the response invalid.

## What it meant

Every response-validation test that checks an error against the spec failed.
That is 127 distinct tests, on both graph backends, across user management,
agents, AI model providers and more. It is the entirety of tonight's nightly
failure, and it is why all four shards went red rather than one.

The failure reads as unrelated to its cause:

```
jsonschema.exceptions.ValidationError:
  Additional properties are not allowed ('requestId' was unexpected)
On instance['error']:
  {'code': 'HTTP_UNAUTHORIZED',
   'message': 'No token provided',
   'requestId': 'it-TestGetAllUsers...'}
```

Nothing about the product is broken here — the responses are fine and the field
is wanted. The spec is simply behind them, and the spec is what the tests hold
the API to.

## The fix, and why this way

The ten error payload schemas in the spec now list `requestId` as an optional
string. No response changes.

The more useful half is the test that comes with it. It sends a real error
through the real error middleware and compares the keys it produced against the
keys the spec allows, in both directions: nothing sent that the spec forbids,
and nothing required by the spec left out. A field added to the error body
without being added to the spec now fails on the pull request that adds it.

That matters because of how this one behaved: it was invisible on the pull
request that introduced it — unit tests pass, the spec is not consulted — and
only surfaced in the nightly hours after it merged, as 127 failures with no
obvious common cause.

## What changes going forward

The error envelope and its published description are checked against each other
on every run.

## How it was checked

The full Node suite passes: 10,375 tests.

The new test was checked against the unpatched spec, where it fails with
`["requestId"]` and points at the schema to update — the same break, caught
where it should have been caught.

## Summary by CodeRabbit

- **Documentation**
  - Updated API error response documentation to include an optional request identifier.
  - Request identifiers can help reference specific requests when reporting issues and may be absent when requests do not reach request processing.

- **Tests**
  - Added validation to ensure error responses match the published API specification.
  - Confirmed that request identifiers are preserved when provided.


shekharkadyan:
On the review point about the test only inspecting `ErrorResponse`: fixed, and it mattered more than it looks. Most of tonight's failures came from the error envelopes routes declare inline, not from the shared schema, so a test that checked only `ErrorResponse` would have passed while the thing that actually broke stayed broken.

The test now walks the whole document for error payloads — any schema that forbids extras and declares `code` and `message` — and checks every one of them, naming each by its path in the file. It reports three kinds of mismatch: a field the schema would reject, something required that is not sent, and a declared type that does not match what is sent. That last one covers `requestId` being declared as anything other than a string.

Against the unpatched spec it now reports the mismatch in all ten schemas instead of one.

shekhar-pipeshub (approved):

